Issue #10, opened 2024-10-24
Apps Design Architecture


rodvar:
Implement foundations to make it easy for developers joining the project to focus on implementation having the basic app layers solved, in this case, separation of concerns domain - flow - presentation.

Bisq2 Desktop uses a modified version of the MVC (Model-View-Controller) design pattern (allows read-only comms from view to model kinda like MVVM does). See [Bisq2 MVC docs](https://github.com/bisq-network/bisq2/blob/main/docs/dev/mvc-model.md) for details.

Since we are building using KMP Compose, views are the Compose views (Composite design pattern based framework views).

This issue can be close donce we have interfaces/abstract classes for each component linking each other and are used in a working example (could be the default barebones view we have now in place)

PR #17, opened 2024-10-27
Feature/app design architecture

rodvar:
- Implementation for https://github.com/bisq-network/bisq-mobile/issues/10
 - MVP (similar to MVC) is being used, BasePresenter implementation provided for each presenter to use
 - added basic lifecycle linking (w/ logging) for every app
 - simplified shared code structure: one shared folder with submodules, starting with domain and presentation for now
 - moved current example code models into domain shared submodule
 - moved example UI code to presentation.ui package


rodvar:
merged straight for productivity reasons (Project stage 0)

bisqapps/docs/ios_to_android_lifecycle_mapping.png
nis-ship-it, line 1: Sorry but I questions as to why have lifecycle on either of them? like on android or ios. I was thinking we were using compose? Sorry I don't the imperative way on both platforms so I'm curious as to why you added, as well the MainPresenter? Is it because of the communication with the androidFullNode?

nis-ship-it: @rodvar

rodvar: hey @nis-ship-it thanks for taking the time to check this out! To the last part of your question, yes we need more flexibility to be able to accomodate for the `androidNode` as well as reuse as much of the code possible for all of the 3 apps.

Regarding the design arch, I'm not sure I understood your questions properly. Regarding the lifecycle, In this PR I provided a link of all the 3 apps lifecycle into the BasePresenter, so every presenter would be able to know if the view is being rendered or not to say it generically speaking. This would be needed regardless of using compose or not, but to that question we do use compose still (check the current view example is App.kt using KMP Compose. Is there a better way? Please share and we can discuss further, thanks!

PR #31, opened 2024-11-08
Added Koin Dependency

nis-ship-it:
Added koin dependency under presentation and domain logic.


bisqapps/shared/presentation/build.gradle.kts
```diff
@@ -87,6 +87,10 @@ kotlin {
             implementation(libs.androidx.lifecycle.runtime.compose)
             implementation(libs.logging.kermit)
             implementation(libs.kotlinx.coroutines)
+
+            implementation(libs.koin.core)
+            implementation(libs.koin.compose)
+            implementation(libs.koin.compose.viewModel)
```
rodvar, line 93: I don't think we need the view model dependency. Please see https://github.com/bisq-network/bisq-mobile/blob/main/bisqapps/README.md#app-architecture-design-choice .
The other 2 are great and will be greatly used for injecting the presenters

rodvar (changes requested):
hey @nis-ship-it thanks for your first contribution!

I've left a comment on the code to review with you.

Also, it would be great if you could also add an example of usage at least of the new libs to be used.

You could for example check how the presenters are being injected in the different main views of each app (e.g. `androidClient` MainActivity) and replace the current hardcoded instanciation of the depedencies (repository, presenter, etc) you could replace it with the Koin DI

That would help a lot for other devs to copy and start using Koin.
Let me know you thoughts!

bisqapps/shared/presentation/build.gradle.kts
nis-ship-it: Added viewmodel because of compose, no worries would demo code

nis-ship-it:
@rodvar I don't know much on ios side like if we can really call kt code? In this case koin

rodvar:
> @rodvar I don't know much on ios side like if we can really call kt code? In this case koin

thanks @nis-ship-it I'll have a look into this soon !

rodvar:
@nis-ship-it had the first look and its looking good, some comments here:

 1. for the naming in iOS (DomainGreetingRepository) -> this is because of the .h file iOS-generated by KMP, it's including the module when you are referring from one module (presentation) to another (domain). Can we investigate using `typealiases` ? Also there is an experimental @ObjCName I tested but couldn't get it to work. We could also accept this different naming in iOS as you did, there shouldn't be much code in swift anyways. 
  2.  the `androidNode` implementation is missing (it would be something identical to what you've done with the `androidClient` I believe.

nis-ship-it:
I tried the aliases but it's the same error so dnt knw about that it would work. Forgot about the node side, I will add koin to it

rodvar:
rebased branch to latest main changes

bisqapps/androidClient/src/androidMain/kotlin/network/bisq/mobile/client/MainApplication.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import org.koin.core.context.startKoin
+
+class MainApplication: Application() {
+    override fun onCreate() {
+        super.onCreate()
+
+        startKoin {
+            androidContext(this@MainApplication)
+            modules(domainModule + presentationModule + androidModule)
```
rodvar, line 17: looking great :)

bisqapps/shared/presentation/build.gradle.kts
```diff
@@ -34,25 +38,22 @@ buildConfig {
...
+    //    buildConfigField("APP_SECRET", "Z3JhZGxlLWphdmEtYnVpbGRjb25maWctcGx1Z2lu")
+    //    buildConfigField<String>("OPTIONAL", null)
+    //    buildConfigField("FEATURE_ENABLED", true)
+    //    buildConfigField("MAGIC_NUMBERS", intArrayOf(1, 2, 3, 4))
+    //    buildConfigField("STRING_LIST", arrayOf("a", "b", "c"))
+    //    buildConfigField("MAP", mapOf("a" to 1, "b" to 2))
+    //    buildConfigField("FILE", File("aFile"))
+    //    buildConfigField("URI", uri("https://example.io"))
+    //    buildConfigField("com.github.gmazzo.buildconfig.demos.kts.SomeData", "DATA",
+    // "SomeData(\"a\", 1)")
```
rodvar, line 50: there are some spacing / indentation changes in your commits, can you review it? we use the default fleet IDE setup

bisqapps/shared/presentation/build.gradle.kts
```diff
@@ -34,25 +38,22 @@ buildConfig {
...
 }
 
 kotlin {
     androidTarget {
-        compilations.all {
-            kotlinOptions {
-                jvmTarget = "1.8"
-            }
-        }
+        compilations.all { kotlinOptions { jvmTarget = "1.8" } }
```
rodvar, line 56: I don't think we want this format change, please review indentation / spacing setup

rodvar (changes requested):
This is pretty much ready to be merge, just some nits in terms of code formatting / unnecessary changes to be reviewed

well done @nis-ship-it !

bisqapps/shared/presentation/build.gradle.kts
nis-ship-it: I'm also using fleet, don't know why this changes came up

nis-ship-it: I know why nvim

rodvar:
Thanks @nis-ship-it for your first contribution to bisq-mobile! :tada: 

Please reach out on matrix if you would like to continue contributing :muscle:

PR #37, opened 2024-11-12
Fix - koin injection under Android Node

nis-ship-it:
Have add koin to all places, I guess?? Have test the 3 clients and works fine. The sample code could be adjusted to your needs, bare that it's just for setting up all project with koin.


nis-ship-it:
@rodvar Really sorry I forgot to tell you to not merge since I knew the androidNode wasn't building on my end, but anyway I have finished that. If you get any error, plz dm me

rodvar:
@nis-ship-it no worries, it was my fault to not to test it in the `androidNode` - we'll learn from this and move on. Thanks for jumping quickly at it!

Issue #6, opened 2024-10-24
Integrate full node POC code into Bisq-mobile project


HenrikJannsen:
https://github.com/HenrikJannsen/andorid_only_poc/tree/use-code-from-production-pr is a plain Android project. We should move the into the new project structure of bisq-mobile.

To consider:

 - protobuf related code should be part of the client code (in this case, part of the `androidNode` module)
 - work on this should start after -> https://github.com/bisq-network/bisq-mobile/issues/5

rodvar:
I will be working on this after setting up the app architecture basic layers

rodvar:
work on this can start now though not merged yet, leveraging this PR -> https://github.com/bisq-network/bisq2/pull/2965

rodvar:
https://github.com/bisq-network/bisq-mobile/pull/29 allows to work on this now.

to generate the jars at the moment please download [this PR branch](https://github.com/bisq-network/bisq2/pull/2965) is PR branch and run `./gradlew clean build publishAll`

rodvar:
https://github.com/bisq-network/bisq2/pull/2965 merged, we can go full speed on this one now

rodvar:
# POC native android bisq2

## Analysis

 - MVController design pattern with services following Bisq2 style being the MainActivity the main view
 - the meat is in the controller which has an init phase that sets everything up for usage (this can correlate directly to our MainPresenter for androidNode)
 - init: setup of jars statics (probably better to strip out of the controller/presenter)
 - init: instanciation of services (we can replace later with DI )
 - initialize: setups all the services needed by the main custom AndroidApplicationService (wraps all the usages of bisq2 jars services :) )
 - currently the POC is **creating a new profile on each session** , there must be some persistance setup work to be done to avoid this
 - the above also prints useful data to screen (like chat messages, connected nodes, etc) all by observing changes to data models
 - all the features implemented in the poc are listed under the comment `// Use cases` which is very handy  for migrating one by one and test through-fully

## Adaptations done

 - Migrated all the POC code to Kotlin which was like a breeze with AndroidStudio code convertion tool
 - Changed some usages of Java futures for Kotlin co routines that gives you more control and are better optimised for mobile env

## Notes for migration and next steps

 - bisq2 jars sync/async is pure java futures. As long as we wrap everything in co-routines I think it should be fine but this needs to be tested through-fully specially on real devices to analyse energy / battery consumption and error handling in case the OS kills anything
[...]

PR #38, opened 2024-11-13
Feature/android node networking POC integration

rodvar:
- Implementation for https://github.com/bisq-network/bisq-mobile/issues/6
 - with this PR the `androidNode` app is in a state where it shows on screen anything that has been developed in shared adding the below:
    - connects to bisq2 network
    - creates a profile if not created and persisted already
    - observes and logs network information
    - listens for private messaging (this is gonna be our trading mechanism!)
    - listens for public messaging (for chat / not included in the MVP but nice to have)
 - to not over complicate things its left like this for further development on architecture that's coming to happen.


HenrikJannsen (approved):
ACK

PR #64, opened 2024-11-19
Feature/lifecycle polishing

rodvar:
- View presenter interface for view accessible lifecycle methods
 - improved base presenter on destroy error handling
 - force overriders of base presenter to call super on key methods with provided impl.
 - BackgroundDispatcher platform-aware to solve lack of support for IO corroutine dispatcher on iOS


rodvar:
presenterlogs

PR #72, opened 2024-11-23
Add offerbook domain

HenrikJannsen:


HenrikJannsen:
Will come with a new PR based on #73.

PR #74, opened 2024-11-25
Add log as extension property

HenrikJannsen:


bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/utils/LogUtil.kt
```diff
@@ -0,0 +1,17 @@
+package network.bisq.mobile.utils
+
+import co.touchlab.kermit.Logger
+
+private val loggerCache = mutableMapOf<String, Logger>()
+
+val Any.log: Logger
```
rodvar, line 7: Instead of extending Any can we add it only to the base classes we are interested in?

Like BasePresenter , maybe BaseService? and any abstract class we are interested in. We probably don't want this dependency to affect the models for example.

rodvar (changes requested):

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/utils/LogUtil.kt
HenrikJannsen: I have some service classes which have no base class... I think logging is useful in many areas at least for dev testing stuff...

rodvar: It is indeed, just don't want the dependency to every single object (that's what happens when you extend Any)

rodvar (changes requested):
@HenrikJannsen I think this one needs more work:

 1. androidNode is not compiling (link to setState() calls in the main service as far as I could check) - this is not being picked up by ci because we have it disabled since we depend on jars)
 2. the change is not working as expected, some longs that should be printing with the presenter name print like `<no-tag>                network.bisq.mobile.client           I  Splash presenter unnattached`
 3.  I really would like to avoid adding the kermit dep to Any object we have in the app, you could refactor to have the reusable code like this
 
 ```
 fun doLoggerGet(anyObj: Any): Logger {
    val tag = anyObj::class.simpleName
    return if (tag != null) {
        loggerCache.getOrPut(tag) { Logger.withTag(Logger.tag) }
    } else {
        // Anonymous classes or lambda expressions do not provide a simpleName
        loggerCache.getOrPut("Default") { Logger }
    }
}
 ```
 
 and then do the kotlin extension on the relevant classes maybe?

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/utils/LogUtil.kt
HenrikJannsen: It is only applied if log is called, no? I don't know the internals of extension functions...

HenrikJannsen:
> @HenrikJannsen I think this one needs more work:
> 
>     1. androidNode is not compiling (link to setState() calls in the main service as far as I could check) - this is not being picked up by ci because we have it disabled since we depend on jars)

Have you latest Bisq 2 republished all libraries? There have been changes as I move state to the base class.

> 
>     2. the change is not working as expected, some longs that should be printing with the presenter name print like `<no-tag>                network.bisq.mobile.client           I  Splash presenter unnattached`
> 

Ups there was `Logger.withTag(Logger.tag)` instead of `Logger.withTag(tag)`

>     3. I really would like to avoid adding the kermit dep to Any object we have in the app, you could refactor to have the reusable code like this
> 
> 
> ```
> fun doLoggerGet(anyObj: Any): Logger {
>    val tag = anyObj::class.simpleName
>    return if (tag != null) {
>        loggerCache.getOrPut(tag) { Logger.withTag(Logger.tag) }
>    } else {
>        // Anonymous classes or lambda expressions do not provide a simpleName
>        loggerCache.getOrPut("Default") { Logger }
>    }
> }
> ```
> 
> and then do the kotlin extension on the relevant classes maybe?

I pushed a commit with an alternative solution. WDYT?

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/utils/LogUtil.kt
rodvar: only if you use the keywords `by lazy` - but anyways your new solution is so much better :guitar:

rodvar (approved):
new solution solves all the issues - ACK

Just not merging it straight because I want to review the rest of the PRs first to decide best merging strategy.

Issue #39, opened 2024-11-13
Shared Repositories: Multiplatform Settings


rodvar:
For the `shared repositories` architecture we will need a at least one way of persisting data. We are unsure if we would need complex data structure, but definitely and easy and fast key-value store will be needed at least for settings.

This issue goals are:

 - Analyse MultiSettings lib for KMP --> https://github.com/russhwolf/multiplatform-settings which looks like the best candidate for a KMP compose project
 - add a shared/data module (or reuse shared/domain probably better, we use a subpackage for data there) and integrate the needed dependencies
 - if needed, provide a wrapper of the lib to make it easy to save key/value (probably not but just mentioning)
 - provide the necessary Koin repository to share this dependency
 - provide an example of usage reading/saving a setting
 - test that it works in the 3 apps.

This dependency will be injected in some of our repositories, best candidate will be SettingsRepository (TBC)

nis-ship-it:
Assign that to me

rodvar:
@nis-ship-it I'm thinking for the work I'm doing [here](https://github.com/bisq-network/bisq-mobile/issues/40) , if your PR includes that the implementation you are building for `Multiplatforms Settings` implements an interface like the following it would be ideal :ok_hand: 

```
// ignore the BaseModel definition I'm working on that in my branch
interface PersistenceSource<T: BaseModel> {
    suspend fun save(item: T)
    suspend fun saveAll(items: List<T>)
    suspend fun get(): T?
    suspend fun getAll(): List<T>
    suspend fun delete(item: T)
    suspend fun deleteAll()
}
```

Let me know how does that sound? :)

rodvar:
I've just implemented the first[ bunch of definitions](https://github.com/bisq-network/bisq-mobile/commit/a7d95408e0a5a2ea6ce445f6a5623f45a1088089), trying to be proactive and share with you so we collide the less possible when merging

nis-ship-it:
@rodvar I'm not getting this part, I was thinking we using the library to store something super small? Why are you having it under shared common module? Plus I don't know about "Shared Repositories" could you elaborate on it?

rodvar:
> @rodvar I'm not getting this part, I was thinking we using the library to store something super small? Why are you having it under shared common module? Plus I don't know about "Shared Repositories" could you elaborate on it?

[...]

PR #76, opened 2024-11-26
Feature/multiplatform settings persistence

rodvar:
- Implementation of https://github.com/bisq-network/bisq-mobile/issues/39
 - key value store common implementation based on most used KMP lib
 - provided common tests with mocked implementation
 - using SettingsRepository for an example of real usage: url and connected flag gets saved now and loaded every time the app starts (xClients)
 - tested on the 3 apps, using keychain for iOS

### DIFF
--- a/bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/model/Settings.kt
+++ b/bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/model/Settings.kt
@@ -7,11 +7,3 @@ open class Settings : BaseModel() {
     open var bisqUrl: String = ""
     open var isConnected: Boolean = false
 }
-
-interface SettingsFactory {
-    fun createSettings(): Settings
-}
-
-class DefaultSettingsFactory : SettingsFactory {
-    override fun createSettings() = Settings()
-}
--- a/bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/model/Settings.kt
+++ b/bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/model/Settings.kt
@@ -1,5 +1,8 @@
 package network.bisq.mobile.domain.data.model
 
+import kotlinx.serialization.Serializable
+
+@Serializable
 open class Settings : BaseModel() {
     open var bisqUrl: String = ""
     open var isConnected: Boolean = false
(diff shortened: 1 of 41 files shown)


bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/model/Settings.kt
```diff
@@ -1,5 +1,8 @@
 package network.bisq.mobile.domain.data.model
 
+import kotlinx.serialization.Serializable
+
+@Serializable
```
HenrikJannsen, line 5: Not form your change, but I would rename `bisqUrl` to a more clear name. Like `bisqApiUrl`

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/persistance/KeyValueStorage.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    override suspend fun save(item: T) {
+        settings[keyPrefix + item.id] = serializer(item)
+    }
+
+    override suspend fun saveAll(items: List<T>) {
+        items.forEach { save(it) }
+    }
+
+    override suspend fun get(id: String?): T? {
```
HenrikJannsen, line 25: Why is id nullable?

HenrikJannsen (commented):
To me it is not clear why an id with null is permitted. Then the prefix is the only map key, no?

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/model/Settings.kt
rodvar: might be a good opportunity to do so, I'll have a look

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/persistance/KeyValueStorage.kt
rodvar: I was using it for a case where you are saving a single obj of a type but now that I think it doesn't really makes sense and could be error-prone, I'll change it soon.

rodvar:

bisqapps/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/persistance/KeyValueStorage.kt
rodvar: found a better solution, is pushed now

rodvar:
rebased branch to latest main changes

PR #83, opened 2024-11-28
AndroidCatHashService

rodvar:
- Implementation for https://github.com/bisq-network/bisq-mobile/issues/58
 - Implemented service dependant on ImageUtils for android node
 - provided unit tests using Roboelectric emulated instrumentation

PR #85, opened 2024-11-29
Add user profile icons to assets

HenrikJannsen:
Bases on https://github.com/bisq-network/bisq-mobile/pull/83
Further PRs are coming. Would be just good to get that merged soon as the images cannot be pushed in one go due Github size restrictions and a rebase does not work then.


rodvar:
Follow up to #83  which is approved.
Merging this one as it contains #83 work

PR #86, opened 2024-11-29
Add user profile icon support for android node

HenrikJannsen:
Based on https://github.com/bisq-network/bisq-mobile/pull/85

Preliminary solution. Follow up PR will come with client and iOS support.

For testing it one need to either delete the data directory in simulator or edit SplashPresenter to not skip user profile creation page.
Only here the (on node) it is applied, not at offers yet.

### DIFF
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/utils/ImageUtils.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/utils/ImageUtils.kt
@@ -13,17 +13,23 @@ import java.io.FileOutputStream
 import java.io.IOException
 
 /**
- * Android images utilitary functions
+ * Android images utility functions
  */
 object ImageUtil : Logging {
-
-    fun composeImage(context: Context, paths: Array<String>, width: Int, height: Int): Bitmap {
+    const val BASE_PATH = "cathash/"
+    fun composeImage(
+        context: Context,
+        basePath: String,
+        paths: Array<String>,
+        width: Int,
+        height: Int
+    ): Bitmap {
         val resultBitmap = Bitmap.createBitmap(width, height, Bitmap.Config.ARGB_8888)
         val canvas = Canvas(resultBitmap)
         val paint = Paint()
 
         paths.forEach { path ->
-            val bitmap = getImageByPath(context, path)
+            val bitmap = getImageByPath(context, basePath, path)
             if (bitmap != null) {
                 val scaledBitmap = Bitmap.createScaledBitmap(bitmap, width, height, true)
                 canvas.drawBitmap(scaledBitmap, 0f, 0f, paint)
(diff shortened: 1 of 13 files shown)


rodvar:
rebased against main

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/utils/ImageUtils.kt
```diff
@@ -53,8 +54,8 @@ object ImageUtil: Logging {
 
     internal fun getImageByPath(context: Context, path: String): Bitmap? {
         return try {
-            log.d { "opening file in path $path"}
-            val inputStream = context.assets.open(path)
+            val fullPath = "cathash/$path"
```
rodvar, line 57: this makes the ImageUtil generic class specific to cat hashing.

Please pass the `cathash` as a baseDir parameter or as part of the path parameter

rodvar (changes requested):
It works beautifully

Only 2 comments on this one, one in the code and this:

 - It would be great if we could use this PR to make the profile icon available for the rest of the app. See screenshots, the profile avatar once landed in home still hard-coded.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/utils/ImageUtils.kt
HenrikJannsen: Its a specific ImageUtils for only the cathash use case. We should rename the class to make that more clear. But yes, we can move it as params.

As there will be bigger changes anyway as we need to generate the images also for client mode I would prefer to skip that for later. Probably that class will move to a KMP compatible version then.

HenrikJannsen:
> It works beautifully
> 
> Only 2 comments on this one, one in the code and this:
> 
>     * It would be great if we could use this PR to make the profile icon available for the rest of the app. See screenshots, the profile avatar once landed in home still hard-coded.

Yes that was my plan, but wanted to get client working first. But spent yesterday the whole day to find a solution for passing the images (as its generated using javafx we cannot simply run that in headless mode, though finally i got a solution for that). But then got aware that it would be much better to do it client side as there can be many images (>1000) and it can cause quite some network load and delay. Will have a look today to port the cathash code from bisq2 to client KMP compatible version. Hope it is manageable.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/utils/ImageUtils.kt
HenrikJannsen: Done

HenrikJannsen:
Rebased

rodvar:
> > It works beautifully
> > Only 2 comments on this one, one in the code and this:
> > ```
> > * It would be great if we could use this PR to make the profile icon available for the rest of the app. See screenshots, the profile avatar once landed in home still hard-coded.
> > ```
> 
> Yes that was my plan, but wanted to get client working first. But spent yesterday the whole day to find a solution for passing the images (as its generated using javafx we cannot simply run that in headless mode, though finally i got a solution for that). But then got aware that it would be much better to do it client side as there can be many images (>1000) and it can cause quite some network load and delay. Will have a look today to port the cathash code from bisq2 to client KMP compatible version. Hope it is manageable.

makes sense, let me know if I can help in any ways

rodvar (approved):
Great work Henrik.
I think we merged this in since is target for node, the request on using the genereated profile in the top-right account avatar can be dealt with in the next PR bringing support for the clients as well :)

Issue #81, opened 2024-11-27
Back button logic


nostrbuddha:
Generalise back button logic in Topbar component, so it works smooth across android/iOS, across RootNav and TabNav

rodvar:
also swipe left-to-right "back gesture" is not working on iOS (the android one: right-to-left does on Android).

nostrbuddha:
Jetpack compose doesn't seems to support the "back gesture" in iOS yet. It's not documented in their official documentation.

But I am inferring this from this discussion @ https://github.com/adrielcafe/voyager/issues/144, where people have created custom solutions to accomplish this.

For us, I created Gesture handler, look for horizontal drag and initiate nav.pop. Made this logic to work only in iOS. So it's working (PR: https://github.com/bisq-network/bisq-mobile/pull/84)
But it doesn't show previous screen preview while transitioning.

Voyager folks implemented that as well. Not voyager's official solution, but people in that issue added a gist.
Understanding that logic and implementing it here will take a while.

Shall I get back to this later, once the other UI flows/issues are done?

rodvar:
closing this one , newer issues will be created when needed

PR #87, opened 2024-11-29
Logging runtime crash fix

rodvar:
- fix runtime crash grabbing logger from views (this will need a refactor, we do not want views logging stuff)


rodvar:
found whilst testing #81

Issue #53, opened 2024-11-15
Background/Foreground Service Setup


rodvar:
Currently we have a POC-based example of how to use bisq core to connect to the network and communicate.
We need to structure that into the Android (node) app to provide the most efficient and reliable service possible.

Goals:

1. Apply Koin DI for all bisq services instances which are mostly singletons, and do the proper configuration of them on the koin module so they are ready to be used (this is important for waking up the app and having all ready when clicking on a notification)
2. Develop an [Android Foreground Service](https://developer.android.com/develop/background-work/services/foreground-services) that shows a silent minimalist notification with the apps service state (e.g. "Bisq is <STATE>") and refactor all the current POC 
3. migrate all the p2p functionality coming from the POC to this service
4. test it runs even after killing the app, notification, on click notification to get back to app etc

following this effort each of the features that uses the shared repositories to communicate wiht the rest of the app should be implemented (see P2P Service preffix coming issues)

Things to discuss and decide with team:

 - app should be persistent? (discuss STOP foreground service button)
 - system notification UI (can't use compose, traditional Android Layout XML)
 - lay down the basic abstractions to allow reusal for the xClients.

rodvar:
on further conversations with the team we identified that also the service will be needed in the `xClients`.

The foreground service approach would be definitly useful in the `androidClient` , the only difference is that it would use a different networking client / services. But the whole notification approach, and low-energy consuming availability non-functional requirements are the same. There is no KMP lib to get the best of both worlds with one single shared code, but we can use expect/actual to code everything (or mostly everything) in kotlin and get the right OS components for iosMain and androidMain.

Probably best to use this issue as the foundational brick for all of this.

rodvar:
Reads for the iOS side:

 - [background modes](https://developer.apple.com/documentation/xcode/adding-capabilities-to-your-app#Enabling-Background-Modes)
 - [BG task scheduler](https://developer.apple.com/documentation/backgroundtasks/bgtaskscheduler)
 - [local notifications](https://developer.apple.com/documentation/usernotifications/scheduling-a-notification-locally-from-your-app)

HenrikJannsen:
Just some update from my side on the network side.

Websockets:
[...]

PR #89, opened 2024-12-03
Feature/resilient background service

rodvar:
**Implementation on both platforms with some limitations described below**

**note** this is a WIP, first PR of a series (even more so for iOS side)

 - Implementation for https://github.com/bisq-network/bisq-mobile/issues/53
 - Registers service into OS enabling comms with the user even if the app is backgrounded/killed
 - Tapping on a notification opens the app, even if the app was killed by the OS
 - Notification permissions requests using latest practices
 - added a sample coroutine that pushes new messages every 10 sec for sample dev purposes
 - notification controller allows the rest of the software components to easily send notifications through the service (just need to inject the controller in your presenter/ service)

** Limitations **

 - on iOS, logs are triggered properly but couldn't actually see any notifications neither in a simulator nor on a real device

### DIFF
--- a/androidClient/src/androidMain/AndroidManifest.xml
+++ b/androidClient/src/androidMain/AndroidManifest.xml
@@ -2,6 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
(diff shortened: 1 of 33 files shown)


rodvar:

rodvar:
rebased against latest main changes

rodvar:
testing code is off by default before merge

PR #92, opened 2024-12-06
Add websocket support

HenrikJannsen:
Currently implemented are those topics:
- MARKET_PRICE (pushes updates for market price data)
- NUM_OFFERS (pushes updates for number of offers per market)
- OFFERS (pushes updates of offers)

We use the same patter as in the 'Easybind' observer library we use in JavaFx where you get the existing data delivered when subscribing the observer. That is very handy as otherwise one need to do the extra call (as needed in default JavaFx observers). So when we subscribe to a topic we get returned the available data as response to our subscribe request.

For MARKET_PRICE and NUM_OFFERS we get the full map of quotes or numOffers per market code. It would be wasteful to do a new subscribe and unsubscribe every time the user changes the currency. So we keep the data in a map and take out the actual number depending on the selected market.

For OFFERS we dont want to get the full data set at each change as that could be a significant data size. 
Instead we use a ModificationType (REPLACE, ADDED, REMOVED) to signal if data was added, removed or if we want to replace the whole data set. The default for the MARKET_PRICE and NUM_OFFERS is REPLACE. For OFFERS we support ADDED/REMOVED to insert items or remove them.
We can send a parameter for subscribing to a topic like currency code, in which case we only would get updates for that currency. If the parameter is not set we get all data. We also handle here the full map of offers per currency, as it would be wasteful to subscribe/unsubscribe at each change of markets.

The payload we get is deserialized to the expected data type at the consumer who knows what data to expect. The generic code does not deserialize that data due lack of type info. Maybe there are some magic hacks to get that resolved as we store the type info in the topic, but I did not get that working so far. 

So adding new topics is just to add the topic with its expected type and the consumer defines what type to expect.
We will see with the next user cases if that covers all we need.

The events are emitted on a single thread executor on the server per subscriber, thus guaranteeing the order, but additionally we use a sequence number to check if the later received events are new as the one we got already.

Rest API calls over Websockets:
To avoid that we need to create new Tor circuits at each rest api call, we use the existing websocket connection to route those http messages to the server and there it will get unpacked to call the local rest api endpoints.
So far most rest api calls have been anyway replaced by the websocket subscribe calls. Only the getMarkets is a GET request to the rest api endpoint (though that could be omitted as the markets are a rather static list and only change if we add a market to the price node).
With POST calls there is still a problem (likely on the server side), so we use the normal http client and not the websocket client for that.

For testing one need to run the HttpApiApp (former we used the RestApiApp, this was renamed to HttpApiApp and covers both RestApi and WebsocketApi).


shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/RequestResponseHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+    Logging {
+    private var requestId: String? = null
+    private var deferredWebSocketResponse: CompletableDeferred<WebSocketResponse>? = null
+    private val mutex = Mutex()
+
+    suspend fun request(
+        webSocketRequest: WebSocketRequest,
+        timeoutMillis: Long = 10_000
+    ): WebSocketResponse? {
+        require(requestId == null) { "RequestResponseHandler is designed to be used only once per request ID" }
```
rodvar, line 33: shouldn't this be done in a mutex.withLock { } context to ensure thread-safety?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/RequestResponseHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+        log.i { "Sending request with ID: ${webSocketRequest.requestId}" }
+        requestId = webSocketRequest.requestId
+        mutex.withLock { deferredWebSocketResponse = CompletableDeferred() }
+
+        try {
+            sendFunction.invoke(webSocketRequest)
+        } catch (e: Exception) {
+            mutex.withLock { deferredWebSocketResponse?.completeExceptionally(e) }
+            throw e
```
rodvar, line 42: please log the exception before rethrow like `log.e { "Failed on invoking callback: ${e.message}" }`

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/RequestResponseHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+    suspend fun onWebSocketResponse(webSocketResponse: WebSocketResponse) {
+        require(webSocketResponse.requestId == requestId) { "Request ID of response does not match our request ID" }
+        log.i { "Received response for request ID: ${webSocketResponse.requestId}" }
+        mutex.withLock { deferredWebSocketResponse?.complete(webSocketResponse) }
+    }
+
+    suspend fun dispose() {
+        log.i { "Disposing request handler for ID: $requestId" }
+        mutex.withLock {
+            deferredWebSocketResponse?.cancel()
```
rodvar, line 59: do we allow disposal whilst a request is being processed? if not please add require(requestId == null) within the mutex context

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/RequestResponseHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+        mutex.withLock { deferredWebSocketResponse = CompletableDeferred() }
+
+        try {
+            sendFunction.invoke(webSocketRequest)
+        } catch (e: Exception) {
+            mutex.withLock { deferredWebSocketResponse?.completeExceptionally(e) }
+            throw e
+        }
+
+        return withTimeout(timeoutMillis) {
```
rodvar, line 45: please handle TimeoutCancelationException and call deferredWebSocketResponse?.completeExceptionally(e) && dispose() to avoid mem-leaks

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/WebSocketClient.kt
```diff
@@ -0,0 +1,183 @@
...
+import network.bisq.mobile.client.websocket.messages.WebSocketRequest
+import network.bisq.mobile.client.websocket.messages.WebSocketResponse
+import network.bisq.mobile.client.websocket.subscription.ModificationType
+import network.bisq.mobile.client.websocket.subscription.Topic
+import network.bisq.mobile.client.websocket.subscription.WebSocketEventObserver
+import network.bisq.mobile.domain.data.BackgroundDispatcher
+import network.bisq.mobile.utils.Logging
+import network.bisq.mobile.utils.createUuid
+
+class WebSocketClient(
```
rodvar, line 35: This is overall a great implementation of a WebSocketClient, quite robust and thread safe.

Just a few optimisations I would suggest to add in:

 1. handle exceptions for `startListening` and `send` calls
 2. Same as in the Handler class, code is needed for handling errors on `request` calls
 3. provide a reconnect method and some way to try automatic reconnect if connection gets lost (This one we can create an issue and do later)
 4. implement unsubscribe to avoid mem leaks (same as number 3)

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/rest_api_proxy/WebSocketRestApiClient.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import network.bisq.mobile.client.websocket.WebSocketClient
+import network.bisq.mobile.client.websocket.messages.WebSocketRestApiRequest
+import network.bisq.mobile.client.websocket.messages.WebSocketRestApiResponse
+import network.bisq.mobile.utils.Logging
+import kotlin.uuid.ExperimentalUuidApi
+import kotlin.uuid.Uuid
+
+class WebSocketRestApiClient(
```
rodvar, line 18: Are we implementing REST protocol over WS ? There is some strict nuances on naming and how to access resources and how to expose them. If we are then great, otherwise let's change the names removing REST prefix please 🙏

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/subscription/ModificationType.kt
```diff
@@ -0,0 +1,10 @@
+package network.bisq.mobile.client.websocket.subscription
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+enum class ModificationType{
```
rodvar, line 6: can we provide a better name for this one? Like DataUpdateRequestType or similar?

shared/domain/src/iosMain/kotlin/network/bisq/mobile/domain/di/ClientModule.ios.kt
```diff
@@ -0,0 +1,16 @@
...
+val iosClientModule = module {
+    single(named("RestApiHost")) { provideRestApiHost() }
+    single(named("WebsocketApiHost")) { provideWebsocketHost() }
+}
+
+fun provideRestApiHost(): String {
+    return "localhost"
+}
+fun provideWebsocketHost(): String {
+    return "localhost" // Default for Android emulator
```
rodvar, line 15: comment is misleading in ClientModule.ios

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/client/ClientMainPresenter.kt
```diff
@@ -1,24 +1,42 @@
...
+        runCatching {
+            coroutineScope.launch { webSocketClient.connect() }
+
+            applicationBootstrapFacade.activate()
+            offerbookServiceFacade.activate()
+            marketPriceServiceFacade.activate()
+        }.onFailure { e ->
+            // TODO give user feedback (we could have a general error screen covering usual
+            //  issues like connection issues and potential solutions)
+            log.e("Error at onViewAttached", e)
```
rodvar, line 31: great comment here, best UX is to inform the user connectivity got lost and automatically try to recover it and announce when we got it back (generally done with a invisible/visble banner under top bar). We'll create an issue to implement this soon.

It would be great for this purpose to have a ping/pong implemented between client and server, so we can calculate latency, inform if there is no response from the server, and things like that.

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/client/ClientMainPresenter.kt
```diff
@@ -1,24 +1,42 @@
...
+        }
     }
 
     override fun onViewUnattaching() {
+        // For Tor we might want to leave it running while in background to avoid delay of re-connect
+        // when going into foreground again.
+        // coroutineScope.launch {  webSocketClient.disconnect() }
```
rodvar, line 38: 👍

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/GettingStartedPresenter.kt
```diff
@@ -1,43 +1,39 @@
...
+    val offersOnline: StateFlow<Int> = _offersOnline
 
     private val _publishedProfiles = MutableStateFlow(1145)
-    override val publishedProfiles: StateFlow<Int> = _publishedProfiles
+    val publishedProfiles: StateFlow<Int> = _publishedProfiles
+
+    val formattedMarketPrice: StateFlow<String> = marketPriceServiceFacade.selectedFormattedMarketPrice
+
+    private var job: Job? = null
+    private val coroutineScope = CoroutineScope(BackgroundDispatcher)
```
rodvar, line 29: please use already available BasePresenter#backgroundScope

rodvar (changes requested):
@HenrikJannsen great PR overall Henrik 💪 . 

I've added these comments on a first pass purely doing code review, I'm gonna do some testing ASAP and put my feedback as well.

This is a foundational PR so we might be a bit more strict on this one I hope you don't hate me for it haha :)

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/user_profile/ClientUserProfileServiceFacade.kt
```diff
@@ -27,58 +25,41 @@ class ClientUserProfileServiceFacade(
...
-        return try {
-            apiGateway.getUserIdentityIds()
-        } catch (e: Exception) {
-            log.e("getUserIdentityIds failed", e)
-            emptyList()
-        }
+        return apiGateway.getUserIdentityIds()
     }
 
     override suspend fun applySelectedUserProfile(result: (String?, String?, String?) -> Unit) {
```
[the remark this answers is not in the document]
HenrikJannsen: You mean instead to return a wrapper object for the data? Or to not return a data at all in suspend functions?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/RequestResponseHandler.kt
HenrikJannsen: Yes, i will move it into the mutex block

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/client/ClientMainPresenter.kt
HenrikJannsen: Yes, the network connectivity needs more work anyway and we need to test all the cases when app gets in background, lose internet connection, gets killed by OS,...
I guess that will be quite some effort and would prefer to delegate it for a later stage. And would like to have Tor support first, as that will be the more problematic mode. With clearnet the connection is very fast established but with Tor it will take a few seconds, so the UX impact need to be handled more specific.

I was reading up about Kotlins exception handling strategy, which differs from Java. From my understanding its best to catch exceptions at the layer where we can do something, which is mostly the user layer. Sometimes we might to be able to try to recover from it by e.g. doing automatic reconnects, etc.... And sometimes its useful for debugging and logging to catch and re-throw/repackage an exception. But I guess its better to keep the code relatively clean from exception handling. In Java the checked exceptions can be quite a pain.... I like Kotlins approach better here.

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/subscription/ModificationType.kt
HenrikJannsen: In JavaFx ListChangeListeners they call it Change (though its a bit diff. design). This is certainly not a good fit here.
It also should not be restricted to a certain collection type. But I guess it will only apply to collections as add/remove only makes sense for those. So maybe CollectionMutationTypr or CollectionModificatinType?

Asked ChatGPT for a good recommendation and it came up with ModificationType as the best choice (i did not mention that term in my query). So maybe its not that bad ;-). I would need to change the bisq2 side as well...

```
Query: What is a good term for the ADD or REMOVE or REPLACE operations on a collection? I pack those operation types in an enum and look for a good enum name .


Reply:
A good term for an enum representing operations like ADD, REMOVE, and REPLACE on collections could be:
Suggestions

    ModificationType (Most Common and Intuitive):
        Describes the purpose clearly: it represents types of modifications you can make to a collection.

ChangeType:

    Concise and easy to understand.
    Reflects the idea of changes applied to the collection.

OperationType:

    Emphasizes the fact that these are specific operations on the collection.

CollectionAction:

    Focuses on the actions performed on a collection.


DeltaType:

    Inspired by the concept of "deltas" (changes) in data structures.
    Works well if your enum is used in contexts where the changes to the collection are tracked over time.

Recommendation

    ModificationType is the most intuitive for general-purpose use cases, as it directly conveys the intent of modifying a collection.
    If your application focuses on change tracking or versioning, ChangeType or DeltaType might be more contextually appropriate.
```

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/user_profile/ClientUserProfileServiceFacade.kt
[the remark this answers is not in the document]
rodvar: Sorry if it was not clear @HenrikJannsen . Is totally ok to return results, what I mean is instead of passing in a callback function parameter as it is right now:

 `applySelectedUserProfile(result: (String?, String?, String?) -> Unit)`
 
 to avoid that call back function and return the values so that the code can grab it like this
 
 `val (value1, value2, value3) = applySelectedUserProfile()`
 
 does that make sense?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/subscription/ModificationType.kt
rodvar: Loved the `CollectionModificationType` idea

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/client/ClientMainPresenter.kt
rodvar: that's correct. I think that core components need to "handle their own concerns" if that makes sense. Obviously if it makes sense to do so (to inform client code) you can rethrow / package and rethrow and exception.
Sounds good Henrik!! For sure we'll test the hell out of it in all apps in all the situations we can think of / recreate.

rodvar:
@HenrikJannsen gonna need some help testing this one, I'll reach out on M

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/subscription/ModificationType.kt
HenrikJannsen: Will consider to rename it, but as I am in the middle of other work, I prefer to do it in a next larger Websocket PR both both Bisq 2 and mobile.

HenrikJannsen:
> @HenrikJannsen gonna need some help testing this one, I'll reach out on M

Added a comment at the initial post about the renamed HttpApiApp in Bisq2.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/domain/user_profile/NodeUserProfileServiceFacade.kt
```diff
@@ -94,9 +94,9 @@ class NodeUserProfileServiceFacade(private val applicationService: AndroidApplic
         return userService.userIdentityService.userIdentities.map { userIdentity -> userIdentity.id }
     }
 
-    override suspend fun applySelectedUserProfile(result: (String?, String?, String?) -> Unit) {
+    override suspend fun applySelectedUserProfile():Triple<String?, String?, String?> {
         val userProfile = getSelectedUserProfile()
-        result(userProfile?.nickName, userProfile?.nym, userProfile?.id)
+        return Triple(userProfile?.nickName, userProfile?.nym, userProfile?.id)
```
rodvar, line 99: that's it! :)

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/websocket/subscription/ModificationType.kt
rodvar: sounds good Henrik

rodvar:
rebased to main updates

rodvar:
android node is working as expected, testing WS clients now

rodvar:
@HenrikJannsen ok awesome, It works on the 3 apps. I'll add some docs on how to setup the env properly so that its easier for everyone to contribute.

There are some things we need to discuss, it looks like each running API instance only supports one client only? This causes issues when using the same API instance with a different device, the app assumes its the other user but obviously the cathash has not been generated in that device and then it won't load properly. We need to think how to solve this and allow support for multiple clients per trusted node ? If this is too complex we can support one for the MVP but soon after work on extending it.

Gonna do the last code review and add docs and merge this one in 💪

rodvar (approved):
Merging as is, please have a look at the comments to discuss/consider for next PR! :) 

The main thing to deal with is support for multiple clients per trusted node.
If its fesiable to do for MVP then great, otherwise we just need to refuse connections from any device that is not the one that generated the profile (and therefore the cathash) for security purposes right? 

I've added some commits with docs, fix to the crash when it can't connect, some logs and the posibility to configure default networking from gradle.properties.

I tested it though in my LAN but it won't work, let's chat about it when you are avail

Awesome work! 💪

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/offerbook/offer/ClientOfferbookListItemService.kt
```diff
@@ -24,48 +28,89 @@ class ClientOfferbookListItemService(private val apiGateway: OfferbookApiGateway
 
     // Misc
     private var job: Job? = null
-    private var polling = Polling(10000) { updateOffers() }
     private var selectedMarket: MarketListItem? = null
```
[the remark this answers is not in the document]
HenrikJannsen: This service is just for the market price. So if we dont want to be triggered by offer market selection but like in bisq2 with a dropdown at the market price, we just change that.

Issue #101, opened 2024-12-10
Hide Notification Bell


rodvar:
For MVP, notifications are minimal (only critical notifications for ongoing trade flow) thus we want to hide the bell until we implement it further down the road.

PR #105, opened 2024-12-11
Hide notification bell for now

rodvar:
- https://github.com/bisq-network/bisq-mobile/issues/101
 - hide notification bell till we implement the whole feature


rodvar:
@nostrbuddha approved?

nostrbuddha:
okay

PR #106, opened 2024-12-11
EnvironmentController implementation

rodvar:
- unified networking config in an environmentController new component that also can provide information of the device
 - provide a fun to detect if running on an emulator


rodvar:
refactor for latest merge #92 + add simulator detection support

PR #113, opened 2024-12-12
Feature/clients bootstrap

rodvar:
- implementation for https://github.com/bisq-network/bisq-mobile/issues/102
 - removed dummy states in favour of real ones
 - provide service to manage websocket. connection in. clients


rodvar:
@nostrbuddha some of the navigation stuff was done but still some missing, seems to be a conflict between the definitions on navigation on start with the client bootstrap facade. We might need to merge both functionalities?

PR #115, opened 2024-12-12
fix/node navigation onboarding

rodvar:
- fix node navigation when profile already exists + improve code readability


rodvar:
contribs to https://github.com/bisq-network/bisq-mobile/pull/115

Issue #51, opened 2024-11-14
Account


nostrbuddha:
10 a) & b)

<img width="529" alt="Screenshot 2024-12-12 at 15 31 33" src="https://github.com/user-attachments/assets/75a7705d-05ef-4ca0-96c6-14c54ec0a237" />

10 c) & d)

<img width="639" alt="Screenshot 2024-12-12 at 15 37 55" src="https://github.com/user-attachments/assets/fdf7c076-8d30-403c-80fa-82f0a8dc1529" />

Refer: https://www.figma.com/design/IPnuicxGKIZXq28gybxOgp/Xchange?node-id=629-14425&t=0NTIOxzkLLAe5cmu-1

Should cover

1. Account - main (10a)
2. Account - User profile (10b)
3. Account - Payment accounts (10c, 10d)

along with their presenter interface definition and its presenter (mocked) implementation (ideally the domain model should be defined and linked with a repository - data on the respository can be mocked of course)

rodvar:
I'm helping @nostrbuddha with 10c as agreed

HenrikJannsen:
Is that required for the MVP?

rodvar:
@HenrikJannsen I do believe for the MVP to work we need to provide the user some sort of Account Payments setup (which is 10c in the design) ? Happy to be proved wrong to slash some stuff from the scope if we can

rodvar:
we are getting block so raising the low-prio bar to start picking this up that will be needed for trading, I'm splitting up work with buddha here

rodvar:
I'm gonna contribute with 10a and 10b screens and functionality

rodvar:
10a/b done

rodvar:
done with the above

Issue #109, opened 2024-12-12
Onboarding: Navigation logic


rodvar:
We now have enough components setup to implement this properly.

Navigation should work as follows.

**A) Fully setup case**

 1. Show Splash Screen with connectivity progress
 2. if any connectivity issue show dialog explaining the problem with a button to open settings, otherwise land on Home

**B) First time use / No profile setup**

 1. Show Splash Screen and go to create profile
 2. after profile is created if any connectivity issue show dialog explaining the problem with a button to open settings, otherwise land on Home
 
 **C) `xClients` - No connectivity details**
 
 1. Show Splash Screen
 2. no trusted no configured found -> go to Trusted node setup screen
 3. proceed as (A) or (B) accordingly.

nostrbuddha:
I like to do this, once I fix the OfferCard layout.

rodvar:
From discussions we had whilst developing settings, we discover the following things that would be good to address in the context of this card:

 1. **Onboarding Setup Screens**: when onboarding redirects the user to a settings screen (e.g. setup trusted node) you shouldn't be able to go back normally as it doesn't make sense to get back to the onboarding setup. We need a consistent way to block the user from normal back navigation and maybe show a dialog to ask if they want to quit?). Ideally this should be bound to BasePresenter so that we can control which presenters block back navigation easily
 2. deep link navigation: not sure if we want to do it from onboarding, but we will need for other use cases (like notifications) a way to deep link the user to nested screen (like in settings). This one is optional and can be strip out into a different issue.

nostrbuddha:
I will do this, this week

rodvar:
@nostrbuddha this has been merged with last Henrik's PR. Just a nit, when nothing is setup (fresh install), the user gets offered a "Create Profile" button, but when clicking it it goes into Trusted Node first.

Text should change based on config, or at least use a generic "Get Started" and then the trusted node app should have a title/description on top explaining why the trusted node setup is needed like "Setup your trusted node ip:port to connect to Bisq network"

PR #117, opened 2024-12-13
Feature/account settings

rodvar:
- 1/2 PRs pf implementation for #51 10a,10b
 - Settings implementation with configurable menu per presenter (allows to configure per app type)
 - Basic implementation with mockup screens
 - navigation of settings fully implemented
 - tested on the 3 apps

### DIFF
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/di/AndroidNodeModule.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/di/AndroidNodeModule.kt
@@ -18,6 +18,7 @@ import network.bisq.mobile.domain.service.user_profile.UserProfileServiceFacade
 import network.bisq.mobile.presentation.MainPresenter
 import network.bisq.mobile.presentation.ui.AppPresenter
 import network.bisq.mobile.presentation.ui.uicases.settings.ISettingsPresenter
+import network.bisq.mobile.presentation.ui.uicases.settings.SettingsPresenter
 import network.bisq.mobile.presentation.ui.uicases.startup.IOnboardingPresenter
 import network.bisq.mobile.presentation.ui.uicases.startup.SplashPresenter
 import org.koin.android.ext.koin.androidContext
@@ -56,7 +57,7 @@ val androidNodeModule = module {
         )
     }
 
-    single { NodeSettingsPresenter(get(), get()) } bind ISettingsPresenter::class
+    single<SettingsPresenter> { NodeSettingsPresenter(get(), get()) } bind ISettingsPresenter::class
 
     single<IOnboardingPresenter> { OnBoardingNodePresenter(get(), get()) } bind IOnboardingPresenter::class
 }
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/di/AndroidNodeModule.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/di/AndroidNodeModule.kt
@@ -6,6 +6,7 @@ import network.bisq.mobile.android.node.domain.market_price.NodeMarketPriceServi
 import network.bisq.mobile.android.node.domain.offerbook.NodeOfferbookServiceFacade
 import network.bisq.mobile.android.node.domain.user_profile.NodeUserProfileServiceFacade
 import network.bisq.mobile.android.node.presentation.NodeMainPresenter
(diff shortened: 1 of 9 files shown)


rodvar:
@nostrbuddha please review and let me know if anything needs to be improved for the coming PR

nostrbuddha:
It's a wow. You created a recursive menu structure for the Settings, with breadcrumb navigation as cherry on the cake.
As done by a true master.

Loved this idea <3

Here is one feedback:

    SettingsMenu, SettingsButton, BreadcrumbNavigation

can go into /components/molecules/*

while /uicases/settings/SettingsScreen.kt will have only 'SettingsScreen'

So my idea throughout the app is having only 'Screen' code inside /uicases, while supporting components goes into atoms/molecules/organisms.

rodvar:
> It's a wow. You created a recursive menu structure for the Settings, with breadcrumb navigation as cherry on the cake. As done by a true master.
> 
> Loved this idea <3
> 
> Here is one feedback:
> 
> ```
> SettingsMenu, SettingsButton, BreadcrumbNavigation
> ```
> 
> can go into /components/molecules/*
> 
> while /uicases/settings/SettingsScreen.kt will have only 'SettingsScreen'
> 
> So my idea throughout the app is having only 'Screen' code inside /uicases, while supporting components goes into atoms/molecules/organisms.

Thanks so much for taking some time to review my PR and for your comments buddha 🙏   Means a lot
I agree with your suggestions, this refactor will come in my next PR 💪 💪

nostrbuddha:
Hi Rodvar,
Actually finding 1 another serious problem with this approach now.

Now each nested menu list is not a separate screen, but rather same screen which renders a portion of the menu Tree.
And each parent to child or going back is not real navigation.

So H/w back button and swipe navigations won't fit right here.

To make this work
 * Having dynamic routes in NavGraph like /settings/{path} and handle with navController.navigate('dynamic-nested-paths')
 * (or) keep current code, handle back button / swipe gestures and simulate the navigation in-screen.

nostrbuddha:
Missed one another point yesterday:

The root component in any screen should be BisqStaticLayout or BisqScrollLayout (If you think you will have much contents, so the screen should be scrollable). They just take care of screen edge paddings.

You can replace the root element `Column` with `BisqStaticLayout`

rodvar:
> Missed one another point yesterday:
> 
> The root component in any screen should be BisqStaticLayout or BisqScrollLayout (If you think you will have much contents, so the screen should be scrollable). They just take care of screen edge paddings.
> 
> You can replace the root element `Column` with `BisqStaticLayout`

thanks buddha this is very useful! - I'll add it in my next PR!

rodvar:
> Hi Rodvar, Actually finding 1 another serious problem with this approach.
> 
> Now each nested menu list is not a separate screen, but rather same screen which renders a portion of the menu Tree. And each parent to child or going back is not real navigation.
> 
> So H/w back button and swipe navigations won't fit right here.
> 
> To make this work
> 
> * Having dynamic routes in NavGraph like /settings/{path} and handle with navController.navigate('dynamic-nested-paths')
> * (or) keep current code, handle back button / swipe gestures and simulate the navigation in-screen.
> 
> For MVP, and since we have minimal screens in settings, I prefer we hard code our screens rather than solving with above solutions.
> 
> Let me know :-)

Absolutely right. As you said, for the particular case of Settings I think we can do this exception but we need to handle the navigation better. Ideally as you said, having some kind of deep navigation 'dynamic-nested-path') solved is ideal because we can redirect users to specific settings screens.

one more thing we need as well talking about this is the possibility of blocking back navigation (probably asking the user if they want to exit the app) in certain screens for onboarding (currently you can go back to the splash/bootstrap screen).

I think I'll create issues (or add into existing) for all of this, and in my next PR for `account settings` I'll do the following:

 - Back button/swipe: if screen is showing root settings, normal behaviour, otherwise go one level up in the settings tree.

I'll let you know how I go with this

rodvar:
@nostrbuddha https://github.com/bisq-network/bisq-mobile/issues/109#issuecomment-2544156572

rodvar:
@nostrbuddha further thinking on this, we don't do back navigation between tab menu items and that makes sense. I think it's understandable from a UX perspective how settings has its breadcrum navigation. The breadcrums let you go back to any point you want. Let's discuss further I'll leave it as is (just added a small change: leaf settings is also added to the breadcrums so its more clear where you are at  :) )

PR #134, opened 2024-12-20
Add i18n support using java property files

HenrikJannsen:
Known issues/missing:
- Not working yet on ios. files need to be added to ios resources and loaded. 
- Processing of the file content need more work (manual linebreaks not detected yet)
- Testing

Loading if the resources takes a few ms. The size is about 260 kb per language. 

We need to put the files in a sub directory in resources as otherwise it would conflict with the resources in node from the i18n jar.

We have for node the resources from the i18n jar already loaded, thus its wasteful to load it again. We could optimize that so that only android client loads the resources. In that case we could remove the mobile subdirectory as the conflict only happens in node. In fact for node we could just delegate the call to the bisq 2 API via Res.

It would be good if we could use only one location where we load the files for ios and android. Not sure if that is possible with the approach to load it from resources. 
If not possible maybe a gradle task takes care that the ios files are always up to date with the android files. Or maybe someone has a better idea....

For the usage I used an extension function on String.
`"myKey".i18n()` or `"myKey".i18n(1, "BTC")` when arguments are passed.

If that style is not welcome feel free to change it. 
In Bisq2 the api is: `Res.get("myKey", 1, "BTC")`


rodvar (changes requested):
Screenshot 2024-12-23 at 11 50 14

for russian chars are not displated properly, probably our current font doesn't support cyrilic ?

I fixed a crash when switching languages, and also detected device language.

I've spent almost 3 hours trying it to get the resources in iOS and couldn't make it (tried from xcode as well). 

I think let's parked this one for now and come back to it later down. the road when we are at the polishing stage for MVP?

HenrikJannsen:
> for russian chars are not displated properly, probably our current font doesn't support cyrilic ? 

Yes, looks like a font issue. We need to have the files in UTF8 encoding then all languages should work.
We have to test the IBMPlexSans if it works well with all languages. I hope so, but in Bisq2 we don't have the more challenging languages like Chinese, Japanese, Vietnamese, Thai,... yet.

> I've spent almost 3 hours trying it to get the resources in iOS and couldn't make it (tried from xcode as well).

Oh, seems to be not that easy. ChatGPT told me its doable and suggested various things but non worked out. 
Yes, lets park it then... I am just a bit concerned that with getting the current use cases (incl. create/take offer) more polished that we need to handle quite a lot of strings and often with parameters, so some work need to be invested which might get later be re-done.

rodvar:
probably good to give priority to this one https://github.com/bisq-network/bisq-mobile/pull/140 when we re-engage with this task

Issue #137, opened 2024-12-20
i18n: provide bisq-compatible solution


rodvar:
To be able to use bisq translation mantainance team we need to draw our i18n in a compatible way

This PR goes in that direction

 - https://github.com/bisq-network/bisq-mobile/pull/134
 
 Though its a POC, the following is needed
 
  - make it work in iOS
  - gradle setup to make everything play nice for builds/releases

rodvar:
done by @HenrikJannsen

PR #140, opened 2024-12-24
I18n support with generated class

HenrikJannsen:
As after several more hours wasted to try to get the resources into iOS, I tried a different approach by creating a gradle task which converts the resources into a kotlin object.

It can be created by:
`./gradlew generateResourceBundles`
and will be located in `build/generated/commonMain/network/bisq/mobile/i18n/GeneratedResourceBundles.kt`
It should be triggered by the relevant build tasks. 

The generated kotlin object though seems to not be accessible in the build as it fails with:
```
Caused by: java.io.FileNotFoundException: /Users/dev/StudioProjects/bisq-mobile/shared/domain/build/tmp/kotlin-classes/debug/network/bisq/mobile/i18n/GeneratedResourceBundles.class (No such file or directory)
```
Maybe @rodvar has an idea what causes that error and how to fix it.

If we get that build error resolved, we could either use that generated kotlin object for all deploy targets or only for iOS and use the normal resources for android.

There would be some headroom for optimisation to only use the selected language to avoid loading all language strings. The generated kotlin object is 1.6 MB, the resources about 2.2 MB.


rodvar:
probably good to give https://github.com/bisq-network/bisq-mobile/pull/140 priority over this one when we reengage with this issue later on

rodvar:
rebased against main

rodvar:
gave a time-boxed effort but couldn't get it to work on iOS yet. needs more work.

rodvar:
this is also trying to solve https://github.com/bisq-network/bisq-mobile/issues/137

Issue #97, opened 2024-12-10
Home: User interactions & data


rodvar:
This issue involve:

 - implementing the observer/observable updates of Market Price (default to USD/BTC), Offers online and published profiles
 - "Enter Bisq Easy" button show navigate to buy/sell tab 
 - "Enter Bisq Easy" -> rename to "Start Trading" (to avoid tech reference on a user communication)
 - "Explore Trade Protocols" -> link should open device preferred browser to our bisq2 wiki corresponding page
 - "Learn More" -> link should open device preferred browser to our bisq2 wiki corresponding page

nostrbuddha:
If you havn't started this already, can you please put it on hold for 1 more day. I have some rough ideas. Will think it through and post here.

rodvar:
@nostrbuddha I barely started so I'll hold to discuss with you :)

nostrbuddha:
<img width="293" alt="Screenshot 2024-12-20 at 10 32 19 AM" src="https://github.com/user-attachments/assets/e0113e04-d60b-4596-9774-98540699261b" />

For androidNode, added 2 extra points to make clear, what this app does. Since we will have 2 apps in playstore.

nostrbuddha:
<img width="297" alt="Screenshot 2024-12-20 at 10 34 00 AM" src="https://github.com/user-attachments/assets/40b874b2-66b1-404b-9040-91dd8a92cce0" />

For xClients, added similar points for clarity.

I am re-thinking the onboarding flow here:

Rather than create profile, setup Trusted node screens,
1. Users jump directly to the home page, after Splash -> Welcome sliders
2. In home page, they read what this 'Trusted node' app is about. So they are also now sure, that they installed the right app from playstore.
3. Then click on 'Setup trusted node' button and connect with remote/friend's bisq instance. At this point, other 2 buttons and all other tab icons are disbaled.
4. Then they go on to do 'Create profile' button.
5. After this, any time they open the app, only "Start trading" button is visible.

In this overall flow, if they close the app in between any of the steps, this home page design, helps them to resume from where they left off.

Let me know, what do you think about this approach?

rodvar:
[...]

PR #143, opened 2024-12-27
Feature/home redesign

rodvar:
- Implementation for #97 
 - provided generic way of navigating to URL using default browser and applied to home navigations
 - Same for blocking UI, with examples on avatar and this home screen. This fixes double-tap UI issues
 - Start Trading takes the user to the Buy/Sell trading screen
 - Applied new desing to home screen
 - fix breadcrum navigation bug on leaf click

### DIFF
--- a/androidClient/src/androidMain/kotlin/network/bisq/mobile/client/di/AndroidClientModule.kt
+++ b/androidClient/src/androidMain/kotlin/network/bisq/mobile/client/di/AndroidClientModule.kt
@@ -1,6 +1,8 @@
 package network.bisq.mobile.client.di
 
 import network.bisq.mobile.client.service.user_profile.ClientCatHashService
+import network.bisq.mobile.domain.AndroidUrlLauncher
+import network.bisq.mobile.domain.UrlLauncher
 import network.bisq.mobile.service.AndroidClientCatHashService
 import org.koin.android.ext.koin.androidContext
 import org.koin.dsl.bind
@@ -8,6 +10,7 @@ import org.koin.dsl.module
 
 
 val androidClientModule = module {
+    single<UrlLauncher> { AndroidUrlLauncher(androidContext()) }
     single {
         val context = androidContext()
         val filesDir = context.filesDir.absolutePath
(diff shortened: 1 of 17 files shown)


nostrbuddha:
One issue. This happens at times, not always.

1. I click on 'Learn more' to open the URL in external browser (Chrome here).
2. Press hardware back button to get back to Bisq app.
3. 'Start trading' is now disbaled.

https://github.com/user-attachments/assets/3e71b523-be78-42a7-a906-1948d86c449a

Tested and happens in both androidNode and androidClient.

nostrbuddha:
1. 'Published Profiles' count is pending from 'domain/services'?

2. `isInteractive` is a great idea :-)
Just a suggestion to improve it further. Rather than enabling/disabling UI controls, the logic can be moved to BisqScroll/StaticLayout, which observes for this state change. And when `isInteractive` is set to true, it render a transparent box over the whole screen, which captures all user input's, thus disabling the controls below. I can do it it next PR. Let me know.

rodvar:
@nostrbuddha buddha thanks for taking the time of reviewing my PR and spotting that bug  - well done 💪 
I'll fix it in a commit in a bit.

Regarding your questions above:

 1. correct
 2. I think that's a great improvement, so we don't have to apply the verification on every single interactive button/link. Happy to get that in in your next PR! We just need to make sure that it only disables the UI whilst processing specific stuff (as bugs like the ones you discovered above are not nice for the UX 😆 ). 
 
 Cheers!

rodvar:
branch rebased

rodvar:
@nostrbuddha please have a look at my commit and review again when you have some time. You helped spotting a major issue that would certainly bring problems in the future so great we tackled it now ;) cheers!

nostrbuddha:
>     2. I think that's a great improvement, so we don't have to apply the verification on every single interactive button/link. Happy to get that in in your next PR! We just need to make sure that it only disables the UI whilst processing specific stuff (as bugs like the ones you discovered above are not nice for the UX 😆 ).

 - Yes. I will do this in the other pending PR, tomorrow night.

PR #144, opened 2024-12-31
Update app images to fit Bisq 2 style and use high resolution

HenrikJannsen:
Did not get the icon with the dark grey background as screen icon working. 
On simulator is shows it with a circular mask and with a light green background around. Not sure if thats only a simulator issue...
On iOS simulator the icon looks as expected.

Also at the app launch the background color if set to back or grey is ignored. When using a color it works but its also masked with a circular mask. Would be nice to have our dark grey as full screen background and the green or light grey icon.

Screenshot 2024-12-31 at 18 31 16


rodvar:
thanks for this @HenrikJannsen it was on my list 🙏 !

rodvar:
I left it as is for iOS though it should be same color scheme as android. App naming to differentiate node from cli. Merging now, if we agree on any changes we can do before release.

HenrikJannsen:
There is not a 100% branding decision regarding Bisq 1 and Bisq 2 logo. But I guess best to stick with current state of Bisq 2. For full node and client node, we also need to figure out how we want to deploy it. I guess best to have only one app and the user can choose at first start which mode to use, and can switch later in settings. But no need to do that now as it will come with some effort to merge the 2 modes for android... To have 2 apps for download might be confusing to users and those who want to use both modes might find it annoying to have to install 2 times.

rodvar:
> There is not a 100% branding decision regarding Bisq 1 and Bisq 2 logo. But I guess best to stick with current state of Bisq 2. For full node and client node, we also need to figure out how we want to deploy it. I guess best to have only one app and the user can choose at first start which mode to use, and can switch later in settings. But no need to do that now as it will come with some effort to merge the 2 modes for android... To have 2 apps for download might be confusing to users and those who want to use both modes might find it annoying to have to install 2 times.

indeed, changing the archticture to one app is not trivial and also throws under the sink lots of flexibility and optimizations by having 2 apps. We decided to have 3 apps (1 node, 2 clients) to give the best solution possible for both, typically something that uses bisq node won't be a user for bisq cli and viceversa so indeed we would be having one bisq app only in the device. But in case there is someone that uses both (like devs ? :) ) then we have then clearly differentiated. 

Happy to change branding colours when we do polishing prior to release

PR #145, opened 2025-01-07
I18n support with statically generated class

HenrikJannsen:
Superseded https://github.com/bisq-network/bisq-mobile/pull/140 and https://github.com/bisq-network/bisq-mobile/pull/134

As using the generated file in the build directory still does not work on iOS, this PR writes the files into the source directory. Further it adds UTF8 support and splits the bundles per language as otherwise the build fails as the file got too large.


rodvar:
solution for https://github.com/bisq-network/bisq-mobile/issues/137

rodvar:
rebased to main

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/bootstrap/ClientApplicationBootstrapFacade.kt
```diff
@@ -36,15 +37,15 @@ class ClientApplicationBootstrapFacade(
             if (!trustedNodeService.isConnected()) {
                 try {
                     trustedNodeService.connect()
-                    setState("Connected to Trusted Node")
+                    setState("bootstrap.connectedToTrustedNode".i18n())
```
rodvar, line 40: @nostrbuddha for the MVP we are ok as we commited support for english-only initially, but right after we need to use this technique "KEY_STRING".i18n() for every text used in the app after this PR gets merged.

rodvar (approved):
Brilliant work @HenrikJannsen 
Just deleted some commented code, merging now.

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/bootstrap/ClientApplicationBootstrapFacade.kt
HenrikJannsen: All mobile only strings should end up in the mobile resource file, and then the generate task should be run to update the kotlin files (only engl. if no translations added). We should stick to a good naming convention to keep the strings well organized with first token as general section (e.g. offerbook) then the subsection (e.g. createOffer),... and we use camelCase in bisq 2 so we should stick to same style. Basically it should follow as close as possible the bisq 2 strings convention and existing keys (e.g. if we only add one new string to an area where bisq2 strings are used we should use the same key prefixes.

HenrikJannsen:
> Brilliant work @HenrikJannsen Just deleted some commented code, merging now.

I left it as one day when we spend more time on it to get it working with files generated to the build directory or even to use the property files in resources directly those gradle snippets might be useful. But its in the history anyway...

rodvar:
> > Brilliant work @HenrikJannsen Just deleted some commented code, merging now.
> 
> I left it as one day when we spend more time on it to get it working with files generated to the build directory or even to use the property files in resources directly those gradle snippets might be useful. But its in the history anyway...

haha yeah indeed that code sounded familiar.. its ok - we are good for now with this improvement

Issue #163, opened 2025-01-15
Networking to use trusted node setup dynamically


rodvar:
Currently the web socket client is using the default gradle setup for ip:port WS connectivity.
We need to make sure that if the trusted node is configured or changed it will pick up the data.

Also implement test connection properly on trusted node setup

rodvar:
done with the above

PR #164, opened 2025-01-15
WebSocketClient dynamic host/port

rodvar:
- Implementation for https://github.com/bisq-network/bisq-mobile/issues/163
 - make dependants on webSocketclient use a provider
 - provider listens to settings repository changes and updates instance cached
 - implementation for test connection on trusted node setup


rodvar:
@nostrbuddha @HenrikJannsen  Also implemented the test connection when setting up trusted node. I'll try to get to the connectivity service and easy "event-driven" emissions on connectivity where every component can attach to to react to connection lost / reconnection

Issue #157, opened 2025-01-12
Auto-reconnect when connection get lost/fails


rodvar:
there are 2 cases here:

 - **Onboarding / app cold start** -> if the connection fails currently the UI just shows a message with no option. We should add a button to retry that attempts reconnection
 - **Whilst on the main view** -> if connectivity gets lost for any reason, it should attempt to reconnect automatically. When #156 is implemented the user will be able to understand this is happening in the background

rodvar:
androidNode handles its own reconnection (bisq2 impl)

PR #165, opened 2025-01-16
Feature/redirect to node setup when no connectivity

rodvar:
- redirect to node instead of trying to fetch profile when there is no connectivity (xclients)
 - code cleanup from previous merge


rodvar:
Partially implements #157

PR #174, opened 2025-01-25
Add num days after redacting trade data to settings

HenrikJannsen:
To sync with https://github.com/bisq-network/bisq2/pull/3134


rodvar:
@HenrikJannsen please review compilation error fix above 🙏

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/mapping/Mappings.kt
```diff
@@ -1120,6 +1120,7 @@ class Mappings {
 
     object SettingsMapping {
         // toPojo method not implemented as we do not have a settings value object in the domain
+        private const val DEFAULT_DAYS_AFTER_REDACTING_TRADE_DATA = 0
```
HenrikJannsen, line 1123: It is 90 days: See: SettingsService.DEFAULT_NUM_DAYS_AFTER_REDACTING_TRADE_DATA = 90;

HenrikJannsen: Also it should use the newly added value not just a default value. On Bisq 2 the new version is in main.

HenrikJannsen (commented):
We should rather use the value form the settingsService and if not yet supported in the rest api layer to add support for it.

rodvar:
> We should rather use the value form the settingsService and if not yet supported in the rest api layer to add support for it.

Hey Henrik @HenrikJannsen , sorry I'm confused with this one!

My contrib was simply to fix the compilation issue (for node). But I actually don't really understand what is this about. 

 - What does "num days after redacting trade data" actually mean?
 - Why do we need that in settings?
 - What would be the use in the mobile apps?

rodvar:
> > > We should rather use the value form the settingsService and if not yet supported in the rest api layer to add support for it.
> > 
> > 
> > Hey Henrik @HenrikJannsen , sorry I'm confused with this one!
> 
> We added the `numDaysAfterRedactingTradeData` field and need to support that in mobile as well. Instead just fixing it with a default value it would be good to just add the missing field in our settings dto. It would be a good and easy exercise hot the back end and the front end play together.
> 
> I will not be available for all those small tasks as I need to move to MuSig soon. So I prefer to focus on the core tasks and leave the rest to you guys.
> 
> > My contrib was simply to fix the compilation issue (for node). But I actually don't really understand what is this about.
> > ```
> > * What does "num days after redacting trade data" actually mean?
> > ```
> 
> It is a privacy feature as we have in Bisq 1 to delete private data of the trade after a certain time. The user can adjust that in the settings. The account data are the primary data we want to redact after some time. Also the trade chat will get deleted. There is an old discussion/issue in Bisq 1 for it... We should add some link to a wiki page to its explained what it means. Not sure if a Wiki page exists for that.

I think part of the job was already done in a partial implementation buddha did [here](https://github.com/nostrbuddha/bisq2/tree/lang_ws)

PR #189, opened 2025-02-05
Update to 2.1.3

HenrikJannsen:


rodvar (approved):
🎉  

Just needed to add a rule for proguard otherwise release build fails. Merging when CI finishes.

PR #191, opened 2025-02-06
Add numDaysAfterRedactingTradeData

rodvar:
follow up of -> https://github.com/bisq-network/bisq-mobile/pull/174

 - [b22df4da6279c402f02cf770f3cd2ae863b2de75] default to bisq core se…ttings value for DEFAULT_NUM_DAYS_AFTER_REDACTING_TRADE_DATA
 - [401fc11ddd802746e3284cebb6f8525757c6581b] fix compilation error of android node
 - [2c231f563595613dc89a982deda084ed961ec98f] remove unused files
 - [75f9c1a422462771cb072db31a9c7f2159c9620b] Add numDaysAfterRedactingTradeData to be in sync with https://github.com/bisq-network/bisq2/pull/3134

### DIFF
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/settings/ClientSettingsServiceFacade.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/settings/ClientSettingsServiceFacade.kt
@@ -87,6 +87,12 @@ class ClientSettingsServiceFacade(val apiGateway: SettingsApiGateway) : Settings
         // Not applicable for xClients
     }
 
+    private val _numDaysAfterRedactingTradeData: MutableStateFlow<Int> = MutableStateFlow(90)
+    override val numDaysAfterRedactingTradeData: StateFlow<Int> get() = _numDaysAfterRedactingTradeData
+    override suspend fun setNumDaysAfterRedactingTradeData(days: Int) {
+        // Not applicable for xClients
+    }
+
     private val _ignoreDiffAdjustmentFromSecManager: MutableStateFlow<Boolean> = MutableStateFlow(false)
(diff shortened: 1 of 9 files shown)


rodvar:
@HenrikJannsen I connected the components with the value coming from the network (node and clients), and made it available for the general settings to display (that's hidden at the moment @nostrbuddha  is working on it). 

Could you please review and let me know if there is a specific application/validation we need to do with that network number?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/settings/ClientSettingsServiceFacade.kt
```diff
@@ -87,6 +87,12 @@ class ClientSettingsServiceFacade(val apiGateway: SettingsApiGateway) : Settings
         // Not applicable for xClients
     }
 
+    private val _numDaysAfterRedactingTradeData: MutableStateFlow<Int> = MutableStateFlow(90)
+    override val numDaysAfterRedactingTradeData: StateFlow<Int> get() = _numDaysAfterRedactingTradeData
+    override suspend fun setNumDaysAfterRedactingTradeData(days: Int) {
+        // Not applicable for xClients
```
HenrikJannsen, line 93: We should implement the API call as with other properties

HenrikJannsen (changes requested):
I think it would be better to fully implement it (incl. UI element and API part).

rodvar:
@HenrikJannsen just checked, this is the UI you are referring to:
Screenshot 2025-02-07 at 17 58 01

I need the settings stuff for that and buddha is working on it, I'll merge this PR and create an [issue](https://github.com/bisq-network/bisq-mobile/issues/200) to do the UI and allow edit (API impl)

HenrikJannsen:
See also the controller in Bisq 2 for the validation of the min/max values.

Issue #130, opened 2024-12-19
Connectivity Service


rodvar:
the goal is to have a ConnectivityService that that:
 
 - allows to verify trusted node connectivity (for `xClients`)
 - allows to verify P2P connectivity (for `androidNode`)
 - allows the rest of the app components to subscribe to get notified about connectivity events (e.g.: CONNECTED, SLOW, DISCONNECTED)

For the trusted node scenario some sort of ping/pong or ACK protocol might need to be implemented

HenrikJannsen:
For the node we have the NetworkService which provides the number of connections and if that would go to 0, its an indication that there is a network problem.

For the clients the webservice would act as connection check, if we get a disconnect there we know we have issues.

rodvar:
> For the node we have the NetworkService which provides the number of connections and if that would go to 0, its an indication that there is a network problem.
> 
> For the clients the webservice would act as connection check, if we get a disconnect there we know we have issues.

Thanks for the comment Henrik, yes this PR considers the webservice connection. I still have the implementation for node pending, your comment is helful to get that finish fast. Is there any way of knowing if the connection is "slow" ? In other words, can we consider the connection to be "slow" if the connection is a single one?

HenrikJannsen:
We can measure the round trip time when doing a request to give an idea about the connection speed.

rodvar:
> We can measure the round trip time when doing a request to give an idea about the connection speed.

yep excellent, is that already in or needs impl? Should I implement some sort of ping/pong same as the trusted node then? 
the other option I can think of is to have an in-memory average trip time that gets updated on every comm on the mobile.

If we go that way we don't even need the ping/pong requests (though is a good practice to implement for WS in mobile given the OS might kill the connection after 30-60 secs of inactivity).

I would go for the ping/pong for the xClients and for the node we can use the measuring option if its complicated to add something like a ping/pong (at probably doesn't make sense either in p2p)

HenrikJannsen:
Just take the duration when we start the request until we get the response. There are many requests, no need to add a dedicated one. We can also accumulate and get means/averages... in Bisq 2 thats done in the connection metrics and displayed in the network UI. But I would consider that all nice to have... the only important is to detect disconnections.

> though is a good practice to implement for WS in mobile given the OS might kill the connection after 30-60 secs of inactivity

There is a disconnect handler. That should be called at any connection disconnect. 

>  if its complicated to add something like a ping/pong 

[...]

PR #198, opened 2025-02-07
Connectivity Service & clients implementation

rodvar:
- implementation for #130 
 - xClients / ws approach is by measuring each webclient request roundtrip time
 - node / p2p approach checks network service connected nodes. Implementation for node will be done in a second PR


rodvar:
Screenshot 2025-02-07 at 14 36 42

transitions (android client)

Issue #156, opened 2025-01-12
Avatar let's the user know connectivity status


rodvar:
- add a subtle dot at the bottom right corned of the avatar
 - dot should be green
   - red: when there is no connectivity
   - yellow: whilst attempting to connect/reconnect
   - green: when user is connected
 - this issue needs #130 to be implemented. UI can be done mocking until its ready or do the whole thing altogether.

rodvar:
@nostrbuddha would you like to take this one?

nostrbuddha:
Yes. Pls assign this to me.

But as we discussed, I can only resume work on the weekend, because of system issues!

rodvar:
> Yes. Pls assign this to me.
> 
> But as we discussed, I can only resume work on the weekend, because of system issues!

Thanks buddha!!! hope that gets sorted soon! 🙏

rodvar:
great work @nostrbuddha

PR #202, opened 2025-02-11
Feature/avatar connectivity status

nostrbuddha:
- https://github.com/bisq-network/bisq-mobile/issues/156

Re-used the image files from bisq2 for the color indicator. Will need some tweaks to the image make it look better.

Tested in all 3 apps for CONNECTED and DISCONNECTED status. SLOW status happened once in a while and it's working


rodvar:

rodvar:
headsup fleet warns me about "incompatible resouces folder", clicking on the button moves the files but obviously it will break the feature.

Screenshot 2025-02-12 at 09 38 38

rodvar (changes requested):
this is great @nostrbuddha !

It's really cool to turn the trusted node on/off and see the connectivity indicator changing :)
Obviously for node its always green since its implementation is pending.

Besides the minor stuff I've contributed, there are some issues we need to address before merging:

 1. If for any reason the default avatar is used (when cat hash fails to be generated, generally in weird reusal of profile scenarios but still can happen) then the semaphore feature is not shown at all
  2. the dots are great but depending on the randomly generated cathash, sometimes they are barely noticeable (sometimes the green one, sometimes the red one). Could you make them render a little bit to the right and bottom of the cathash instead of overlapping?

nostrbuddha:
> headsup fleet warns me about "incompatible resouces folder", clicking on the button moves the files but obviously it will break the feature.
> Screenshot 2025-02-12 at 09 38 38

Yes. This happens, since we have sub-folders inside /drawable. Jetpack expects all assets files to be directly under /drawable. But sub-folders works with our DynamicImage.

rodvar (approved):
Absolutely stunning @nostrbuddha 

Screenshot 2025-02-12 at 15 11 46
Screenshot 2025-02-12 at 15 10 19
Screenshot 2025-02-12 at 15 08 13
Screenshot 2025-02-12 at 15 08 02
Screenshot 2025-02-12 at 09 52 27


I'm ignoring fleet warning as you suggested we can deal with that later if it proves to bring issues (e.g. for release)

Issue #129, opened 2024-12-19
Trading: easy clipboard copy/paste functionality


rodvar:
This issue is about verifying that clipboard copy/paste works well in both platforms and each app and adding native support for it if not so that it simplifies trading experience copying large hash strings (btc addresses, lightning invoices etc)

nostrbuddha:
Assign this to me please.

PR #206, opened 2025-02-12
Copy/paste functionality in Trade flow

nostrbuddha:
- Implementation for #129 
 - Copy buttons were already added in some previous PRs. Missed few places, added them now in this PR
 - Paste button added now.
 - In bisq2, same icon is used for both copy/paste. But here, I added different icons.


rodvar (approved):
works great, good adaptation of UI component for reusal 💪

Issue #146, opened 2025-01-07
Back Navigation Gestures


rodvar:
Make sure back navigation works (left to right swipe in iOS) and the implementation suggested/commented `BackHandler.ios.kt` is the one to obtain same navigation results as android:

 - finishing/cancelling workflow removes it from the "back stack" and gets you back to the original screen where you started.
  - Tab home navigation: back shows you warning of exit and back again does default OS behaviour
  - other tabs back navigation: get back to home tab

rodvar:
good news is, the effort we did in the past to bring back navigation gestures in ios (left->right swipe) is working, but the actual implementation of back is inconsistent comparing to android (more mature to what we are looking for).

So the effort to finalise it is in bringing the same implementation. 

Also need to check if this bug is happening also in the rest of the apps: when onboarding, back swipe should quit the app (do not allow to flow back to a setup screen that has been already completed).

rodvar:
as per ☝

PR #210, opened 2025-02-13
several navigation fixes

rodvar:
- Implementation for #146 
 - now swipe works identically as in Android except when the user is already on home (for iOS there is no actual "home button" concept as in android, users swipe down->up to put the app in the background and that is covered!)
 - TODO complete me further

### DIFF
--- a/shared/presentation/src/androidMain/kotlin/network/bisq/mobile/presentation/PlatformPresentationAbstractions.android.kt
+++ b/shared/presentation/src/androidMain/kotlin/network/bisq/mobile/presentation/PlatformPresentationAbstractions.android.kt
@@ -11,3 +11,7 @@ actual fun getPlatformPainter(platformImage: PlatformImage): Painter {
 }
 
 actual fun getPlatformCurrentTimeProvider(): TimeProvider = AndroidCurrentTimeProvider()
+
+actual fun exitApp() {
+    // not used in Android
+}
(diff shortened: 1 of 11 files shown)


rodvar:
@nostrbuddha is ready now, please review when you have a moment 🙏

nostrbuddha:
Yes. Will check this later today.

nostrbuddha:
1. Navigate to a different tab.
2. Swipe back, It goes back to home screen. (As expected)
3. Do swipe back once in home screen, to get the "Swipe one more time to exit" snackbar.
4. Now navigate to a different tab.
5. Swipe back and the app is closed (It should nav back to home screen, as it did previously)

https://github.com/user-attachments/assets/8bb02b94-1008-4efd-8067-3956b53912be

Guess it's not done with iOS, as you noted in the comments. Swiping back from home tab in iOS does nothing.

rodvar:
> 1. Navigate to a different tab.
> 2. Swipe back, It goes back to home screen. (As expected)
> 3. Do swipe back once in home screen, to get the "Swipe one more time to exit" snackbar.
> 4. Now navigate to a different tab.
> 5. Swipe back and the app is closed (It should nav back to home screen, as it did previously)
> 
>  AndroidBack.mov 
> Guess it's not done with iOS, as you noted in the comments. Swiping back from home tab in iOS does nothing.

thanks buddha! I forgot to mention this. This bug was not introduced in the PR though, but we need to trace it and fix eventually. I will use your comment above to create an issue and try to tackle it today!

Issue #192, opened 2025-02-06
User gets warned before eliminating own offers


rodvar:
Currently after creating an offer in a certain market, e.g. Euro, you can see it immediately in the offerbook Euro offers.

If you tap your own offer it gets removed. Let's add a confirmation dialog for it "Would you like to remove your offer? Remove offer / Cancel"

PR #211, opened 2025-02-14
warn user before delete own offer

rodvar:
- Implementation for #192 
 - implement warning dialog before deleting self offer and error message comm via snackbar/toast


rodvar:
@nostrbuddha please review 🙏

nostrbuddha:
This is good.

Just a thought.
Do we need to make the entire card clickable to show delete action (as of now)? (or) will it be better to restrict only to the `delete` icon at the right edge?

rodvar:
> This is good.
> 
> Just a thought. Do we need to make the entire card clickable to show delete action (as of now)? (or) will it be better to restrict only to the `delete` icon at the right edge?

Good point. If the only action you can do with it its delete its OK (now that it has a warning)

But eventually if we add a more detailed view of the offer then the expectation would be that clicking the card takes you there and the delete button is just the icon so if you would like to change it in your next PR its great

nostrbuddha:
Sure. Will do.

Issue #194, opened 2025-02-06
Create Offer: Edit amount manually does not update the amount


rodvar:
1. open the app
 2. go to offerbook (buy/sell tab)
 3. tap on the "+" icon to create an offer
 4. choose buy and click on next
 5. select any currency
 6. move slider and observe -> the amount changes based on the slider position
 7. now tap on the amount and try to edit manually

ER: either the edit should be blocked or if allowed it should edit immediately and also move the slider automatically accordingly
OR: edit is permitted but trying to change the numbers does nothing

PR #216, opened 2025-02-17
Do not allow manual edition of amount on create fixed offer

rodvar:
- Fix for https://github.com/bisq-network/bisq-mobile/issues/194


rodvar:
@nostrbuddha please review 🙏

Issue #215, opened 2025-02-17
Profile Settings: profile age always shows "less than a day"


rodvar:
even for long used profiles.

rodvar:
Confirmed that the value is coming as `now()` from domain.

Inspecting code in bisq2 it looks like the every time there is a profile update  (e.g. auth) that `publishDate` gets updated so the publish date is not the publication date but rather the latest update timestamp and we can't use it to determine profile age

rodvar:
there is a bisq2 ProfileAgeService that has this method

```
public Optional<Long> getProfileAge(UserProfile userProfile) {
        return Optional.ofNullable(dataSetByHash.get(userProfile.getProfileAgeKey()))
                .flatMap(e -> e.stream().findFirst()) // We have only 1 item in the dataSet
                .map(AuthorizedTimestampData::getDate);
    }
```

ProfileCardDetailsController uses the ReputationService to get that service and use that method

rodvar:
@HenrikJannsen is taking this one

rodvar:
moved to node milestone to close connect one

rodvar:
investigate this to help us test the wild wild west of android devices -> https://www.browserstack.com/open-source

PR #217, opened 2025-02-18
profile age

rodvar:
- implementation for #215 
 - publication date handled in mobile as publishDate coming from p2p …profile info is rather the last activity date and is always the `today`


rodvar:
@HenrikJannsen proposed fix, please review

HenrikJannsen:
Profile age and last activity data are provided by the network

rodvar:
> Profile age and last activity data are provided by the network

I'll drop this PR then an investigate how to make it available for mobile

Issue #185, opened 2025-02-04
UI Polishing


rodvar:
Prior to release we need to fix all the small nuances we left aside during the complex functionality implementation on development. 
This is normally referred by the team as the "polishing stage"

# General guidelines

 - **Maintain a consistent communication**. E.g. same type of buttons should have the same size, spacing, padding ,etc. Look the same so the user understand is our app action button without even thinking. Same goes for every UI component
 - Make sure every asyncs interaction blocks the UI (to avoid double requests , there is a handy isInteractive flag avail in BasePresenter) and also uses the right spinner/loading component.
 - obviously any hardcoded data to showcase UI should be eliminated, and if we don't have domain data to show something `simply hide it or show N/A` (e.g. reputation)

# Specific comments per screen / UI component

## Offerbook (market list)
 
 1. Add Button (+) -> 2 issues, should be more rounded (circle like) and the "+" is not very visible, probably good to replace with a + white asset

## Offers (selected market)

 1. Add Button (+) is missing (it should show the workflow skipping the "select currency" because its already selected)
 2. **bug**: **Sell to / Buy From** selector overlaps end of the component when buy from is selected
 3. **bug**: sometimes the "Sell to/ Buy From" selector is unresponsive - hi prio issue causes frustration on users

## Offer card

 1. **bug**: misplacement of elements, language should be anchored to bottom left w/ padding,  payment methods should be anchored to bottom right w/ padding, overall spacing is not consistent across offers
 2. reputation is hardcoded - need to reflect the real offer data .

## Create Offer

 - how To build reputation?: **bug** `Learn how to build reputation` button does nothing, should open browser to corresponding bisq2 wiki
 - Amount: Several minor UI issues with paddings (subtitles too compressed), selector not covering "Fixed amount" text when selected, 
 - Amount: Range amount gets out of aligment with the most inflationary currencies (because they need more text and the space is not fixated). It should have a fix space and reduce the font size to make it fit
 - Currency:  **bug** Search does not work (cannot write anything in the search box)

## Trade Flow: payment method

 - Fiat transfer currency text is hardcoded to USD, should use the selected currency by the user
 - pre title "Step x/3" and title spacing too shallow

## Trade Flow: review trade

 - confirm button significantly larger than back button.

## Trade Flow: Trade ID initial screen

 - reputation is hardcoded to 3.5 stars

## Trade Flow: Trade completed

 - "export trade data" does nothing. Hide it and create a feature for next milestone?
[...]

Issue #116, opened 2024-12-13
Applications default Font


rodvar:
Analyse the possibility of using Bisq2 same font

```
@font-face {
    font-family: "IBM Plex Sans";
    src: url("/fonts/IBMPlexSans-Regular.ttf");
}

@font-face {
    font-family: "IBM Plex Sans Bold";
    src: url("/fonts/IBMPlexSans-Bold.ttf");
}

@font-face {
    font-family: "IBM Plex Sans SemiBold";
    src: url("/fonts/IBMPlexSans-SemiBold.ttf");
}

@font-face {
    font-family: "IBM Plex Sans Medium";
    src: url("/fonts/IBMPlexSans-Medium.ttf");
}

@font-face {
    font-family: "IBM Plex Sans Light";
    src: url("/fonts/IBMPlexSans-Light.ttf");
}

@font-face {
    font-family: "IBM Plex Sans Thin";
    src: url("/fonts/IBMPlexSans-Thin.ttf");
}

@font-face {
    font-family: "IBM Plex Mono";
    src: url("/fonts/IBMPlexMono-Regular.ttf");
}

```

rodvar:
@nostrbuddha  is working on this

PR #218, opened 2025-02-18
Update/theme matching

nostrbuddha:
- Increase font size by 2pt. across the screens.
 - Commented color usage with bisq2 colors (marked the css class name used in bisq2 in BisqTheme.colors for easy identification)
 -  BG color of controls is defined to be a relative color `rgba(255, 255, 255, 0.05)` (https://github.com/bisq-network/bisq2/blob/38321991a10c6807a81d0ca83701525d330ca9fb/apps/desktop/desktop/src/main/resources/css/text.css#L620). Did the same in mobile now. Defined as `secondary = Color(0x0DFFFFFF)`. But I prefer using a absolute color without transparency. @rodvar what do you think?
 - BisqButton already have Default, Outline, Clear types. There are many more styled defined in bisq2. But I am not sure if we are using them in mobile now. So didn't copy the other styles, which are not used here.
 - Added ReportBugPanel that matches bisq2 'Report bug popup' and used it in place of `GenericErrorPanel` in `ErrorOverlay` (If this is good, can rename this back to `GenericErrorPanel`)


rodvar (changes requested):
Overall, UI looks cleaner with all the spacing fixes which contributes to #185  👍 
and also I understand this kills #116 right

I think we need to address some stuff before merging though:

 1. text size increase: I think is too big now by default, for example with default android accessibility settings it looks as shown (see screenshots). I would revert those changes
 2.. I didn't understand the name change on GenericError, cause the current usage seems more accurate to the original naming
 3. I do agree with your color preference. However, I must confess the change does make it look more bisq2 style
  
  
Screenshot 2025-02-18 at 17 54 38

  
Screenshot 2025-02-18 at 17 53 39

nostrbuddha:
Thanks for the feedback, Rodvar.

 1. Forgot to mention about #116. we were using `IBM Plex sans` from the beginning. I just copied the TTF files again from bisq2 repository. To test, I substituted the ttf files with a different weird looking font-family to check, if all texts in app are affected.
 2. Regarding font size increase, can you please tell me how you feel (bigger fonts) with other screens as well? I think home page texts are big which needs to be reduced, but in other screens I think this bigger font looks good. But let me know, if you feel otherwise, I will revert.
 3. Replaced `GenericErrorPanel` with `ReportBugPanel`, which has following look, as in bisq2. On clicking report, it copies the `error text` to clipboard and opens bisq-mobile's github issue page.
Screenshot 2025-02-18 at 3 48 01 PM
 4. So shall I try replacing `secondary = Color(0x0DFFFFFF)` with similar looking absolute color, without transparency. okay?

rodvar:
> Thanks for the feedback, Rodvar.
> 
> 1. Forgot to mention about [Applications default Font #116](https://github.com/bisq-network/bisq-mobile/issues/116). we were using `IBM Plex sans` from the beginning. I just copied the TTF files again from bisq2 repository. To test, I substituted the ttf files with a different weird looking font-family to check, if all texts in app are affected.
> 2. Regarding font size increase, can you please tell me how you feel (bigger fonts) with other screens as well? I think home page texts are big which needs to be reduced, but in other screens I think this bigger font looks good. But let me know, if you feel otherwise, I will revert.
> 3. Replaced `GenericErrorPanel` with `ReportBugPanel`, which has following look, as in bisq2. On clicking report, it copies the `error text` to clipboard and opens bisq-mobile's github issue page.
>    Screenshot 2025-02-18 at 3 48 01 PM
> 4. So shall I try replacing `secondary = Color(0x0DFFFFFF)` with similar looking absolute color, without transparency. okay?

thanks for the clarifications Buddha!

 1. 👍 
 2. see screenshot below. it does look more readable, brings more known-issues on UI components growing out of size and pushing each other though. Maybe try reducing the text size in the home description and fix the ui component issues in this PR context?
 3. I see now. Since we have this ready, should we link it to observing `MainPresenter#_systemCrashed` ? It's all bind, just need to implement what to do when its value changes :) 
 4. happy for you to try this

Screenshot 2025-02-19 at 10 28 39

nostrbuddha:
Sure Thanks. Will fix things and push.

nostrbuddha:
> I see now. Since we have this ready, should we link it to observing MainPresenter#_systemCrashed ? It's all bind, just need to implement what to do when its value changes :)

So here is what is done:
 - Both `presenter.genericErrorMessage` and `MainPresenter._systemCrashed` are listened to.
 - When `genericErrorMessage` got some value and `_systemCrashed is false`, ReportBugPanel is shown. User is allowed to close the popup and continue using the app.
 - But when `genericErrorMessage` got some value and `_systemCrashed is true`, ReportBugPanel is shown. Now user can either close the app or report the issue. Can't continue to use the app.

Is this good?

rodvar:
> > I see now. Since we have this ready, should we link it to observing MainPresenter#_systemCrashed ? It's all bind, just need to implement what to do when its value changes :)
> 
> So here is what is done:
> 
> * Both `presenter.genericErrorMessage` and `MainPresenter._systemCrashed` are listened to.
> * When `genericErrorMessage` got some value and `_systemCrashed is false`, ReportBugPanel is shown. User is allowed to close the popup and continue using the app. (non critical errors)
> * But when `genericErrorMessage` got some value and `_systemCrashed is true`, ReportBugPanel is shown. Now user can either close the app or report the issue. Can't continue to use the app.
> 
> Is this good?

excellent, I'm testing it right now!

rodvar (approved):
great Job Buddha!

Only nit is that both iOS and Android, the confirmation of the trade text has too much opacity its barely readable. We can fix that on the polishing issue context, I'll add a note in

rodvar:
> great Job Buddha!
> 
> Only nit is that both iOS and Android, the confirmation of the trade text has too much opacity its barely readable. We can fix that on the polishing issue context, I'll add a note in

Screenshot 2025-02-19 at 15 08 16
Screenshot 2025-02-19 at 15 04 26

PR #224, opened 2025-03-04
Connectivity service is connected for node

rodvar:
- Implementation for node #130 
 - only pending is usage of round trip log method on node messaging to track isSlow (low prio)
 - connectivity service "isConnected" implementation for node by keep…ing track of connections/disconnections to the node

### DIFF
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
@@ -1,18 +1,57 @@
 package network.bisq.mobile.android.node.service.network
 
+import bisq.common.network.TransportType
+import bisq.network.identity.NetworkId
+import bisq.network.p2p.message.EnvelopePayloadMessage
+import bisq.network.p2p.node.CloseReason
+import bisq.network.p2p.node.Connection
+import bisq.network.p2p.node.Node
 import network.bisq.mobile.android.node.AndroidApplicationService
 import network.bisq.mobile.domain.service.network.ConnectivityService
 
 class NodeConnectivityService(private val applicationService: AndroidApplicationService.Provider): ConnectivityService() {
+    // TODO ConnectivityService#newRequestRoundTripTime() call needs to be applied when a P2P roundtrip call is done for the parent isSlow impl to work
+
+    private var connections = 0
+    private val nodeListener : Node.Listener = object: Node.Listener {
+            override fun onMessage(epm: EnvelopePayloadMessage?, conn: Connection?, network: NetworkId?) {
+                // do nth
+            }
+
+            override fun onConnection(connection: Connection?) {
+                log.d { "connection stablished $connection" }
+                connections++
+            }
+
+            override fun onDisconnect(connection: Connection?, closeReason: CloseReason?) {
+                if (connections > 0) {
+                    log.d { "connection lost $connection for $closeReason" }
+                    connections--
+                }
+            }
+
+        }
+
+    override fun onStart() {
+        try {
+            val serviceNode = applicationService.networkService.get().findServiceNode(TransportType.CLEAR).get()
+            serviceNode.nodesById.addNodeListener(nodeListener)
+        } catch (e: Exception) {
+            log.w(e) { "failed to start node monitoring for connectivity service" }
+        }
+    }
+
+    override fun onStop() {
+        try {
+            val serviceNode = applicationService.networkService.get().findServiceNode(TransportType.CLEAR).get()
+            serviceNode.nodesById.removeNodeListener(nodeListener)
+        } catch (e: Exception) {
+            log.w(e) { "failed to start node monitoring for connectivity service" }
+        }
+    }
+
     override fun isConnected(): Boolean {
-        // TODO implement this and also the ConnectivityService#newRequestRoundTripTime() call needs to be applied when a P2P roundtrip call is done
-        return true
-//        return try {
-//            log.d { "Connected nodes = ${applicationService.networkService.get().numberOfConnectedNodes}"}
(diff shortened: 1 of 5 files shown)


androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
```diff
@@ -1,18 +1,57 @@
...
+                    log.d { "connection lost $connection for $closeReason" }
+                    connections--
+                }
+            }
+
+        }
+
+    override fun onStart() {
+        try {
+            val serviceNode = applicationService.networkService.get().findServiceNode(TransportType.CLEAR).get()
```
HenrikJannsen, line 37: We should support Tor as well. Best to iterate over all and take first found. We do not expect the case to support multiple transports for mobile.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
```diff
@@ -1,18 +1,57 @@
...
+            serviceNode.nodesById.addNodeListener(nodeListener)
```
HenrikJannsen, line 38: Call the listener code (refactor to method) as with addNodeListener we only get the listener called at updates after that. But we want to set the local state also at onStart.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
```diff
@@ -1,18 +1,57 @@
...
+        } catch (e: Exception) {
+            log.w(e) { "failed to start node monitoring for connectivity service" }
+        }
+    }
+
+    override fun onStop() {
+        try {
+            val serviceNode = applicationService.networkService.get().findServiceNode(TransportType.CLEAR).get()
+            serviceNode.nodesById.removeNodeListener(nodeListener)
```
HenrikJannsen, line 47: We use the defaultNode in Bisq 2 (NetworkInfo) not nodesById. nodesById include the nodes for private messages. defaultNode is the base node for all the connection handling. If that has no connections anymore we have troubles even if there would be a userprofile specific node connected.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
```diff
@@ -1,18 +1,57 @@
...
+
+    private var connections = 0
+    private val nodeListener : Node.Listener = object: Node.Listener {
+            override fun onMessage(epm: EnvelopePayloadMessage?, conn: Connection?, network: NetworkId?) {
+                // do nth
+            }
+
+            override fun onConnection(connection: Connection?) {
+                log.d { "connection stablished $connection" }
+                connections++
```
HenrikJannsen, line 23: Better use the  `node.getNumConnections()` as done in `onNumConnectionsChanged` in Bisq 2. Avoid replicating data as it comes with risk of out of sync states.

HenrikJannsen (changes requested):

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
rodvar: good idea to leave that already done, will do on next PR

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
rodvar: sorry I don't quite understand this comment, you mean that nodeListener val should be defined also in onStart() ?

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
rodvar: thanks for the correction Henrik, will apply in next PR

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
rodvar: I tried using a call to node connections first but couldn't get it to work (wan't updating). I'll give a go to your comment and apply in next PR, thanks.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
rodvar: defaultNode related code seems to have been left abandoned, the connections UI related code is actually using

`model.getServiceNode().getNodesById()` as this PR does and it works.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
rodvar: I've just tested and couldn't get it to work, if we do something like

```
override fun isConnected(): Boolean {
        var connections = 0
        findAllServiceNodes().forEach {
            connections += it.peerGroupManager.get().node.numConnections
        }
        log.d { "Connected peers = $connections"}
        return connections > 0
    }
```

it will always return true (connections == 3, one node per network type , probably the service one?) 

please let me know if you see any obvious error?

rodvar: sorted out

PR #228, opened 2025-03-06
Trade guide, Wallet guide, Link buttons

nostrbuddha:
- Trade guide
 - Wallet guide
 - Link buttons with open link confirmation popup
 - RootNavGraph - Code optimization


rodvar (changes requested):
We need to work on some issues for this one Buddha, I'll try to describe below:

1. "Trade Guide": last step "next" button for android still says "next", when clicking it either nothing happens or a white screen is shown
2. "Trade Guide" (iOS): "next" button is greyed out, only way out is navigating back (5 times). Probably better to show "Close" and close the whole guide workflow for all platforms?
3. "Open web link" dialog shows the url between `` which is confusing and looks weird, probably better to simple show the url in bold ?
4. probability hint for offer has missing replacements "Your offer has {0} chances to be taken at this price. Why?"

that's so far what I found, I'll post some screenshots for you now

rodvar:
Screenshot 2025-03-06 at 14 59 49
Screenshot 2025-03-06 at 15 02 25
Screenshot 2025-03-06 at 15 03 26
Screenshot 2025-03-06 at 15 07 24

nostrbuddha:
okay.
The last screen in Trade guide, usually requires the 'I understand' at the bottom of the screen to be checked. So the 'next' button gets enabled. But this white screen is something I didn't notice before.
Will check and fix later today.

rodvar:
> okay. The last screen in Trade guide, usually requires the 'I understand' at the bottom of the screen to be checked. So the 'next' button gets enabled. But this white screen is something I didn't notice before. Will check and fix later today.

for the "accept terms" is hidden inside the scroll view. Maybe we should add a link on top of the action buttons to "Scroll to bottom" that dissapears after click leaving the user ready toaccept and finish?

rodvar:
rebased against main branch

rodvar:
@nostrbuddha the link to bottom button is beautiful. Let's merged this as is and we can do the nits in the next PR ;)

PR #234, opened 2025-03-13
Removed lyricist and replaced with i18n()

nostrbuddha:
- Replaced all lyricist with i18n() calls. Where ever the language key/value is not present (mobile specific strings), added the comment `TODO:i18n`
 - Removed lyricist library dependency
 - Cleanup imports, BisqText calls


rodvar (approved):
cool @nostrbuddha

Issue #238, opened 2025-03-14
Payment accounts: Functional integration


nostrbuddha:
- Payment accounts persistance for androidNode and xClients
 - WS layer in bisq2 for Payment accounts
 - Include Payment account selection in Trade flow

As per discussion with Henrick, implement only UserDefinedFiatAccount and UserDefinedFiatAccountPayload. Ignore the Account hierarchy of classes as it is now in Bisq2 code and mention it in a comment

PR #245, opened 2025-03-20
Update/payment accounts

nostrbuddha:
PR for https://github.com/bisq-network/bisq-mobile/issues/238

 - Payment account CRUD for both androidNode, xClients
 - Payment account dropdown in Tradeflow

Only UserDefinedFiatAccount and UserDefinedFiatAccountPayload are handled. Account hierarchy of classes, as in Bisq2 code in ignored now for bisq-mobile

To be merged along with the Bisq2 API PR @ https://github.com/bisq-network/bisq2/pull/3306


rodvar (changes requested):
haven't reviewed the code yet because functionality wise it looks like it needs more work

 1. node: not working - after entering the inputs and saving the account nothing seems to be happening
 2. connect apps: create account works fine, but generates some visual issues (see screenshot, button positioning, text replacement). Delete sometimes fails and no error is presented to the user.
 
 Also the UI is a bit confusing it feels like one account overlaps the other instead of seeing a list of the available accounts. 
 
 
 
Screenshot 2025-03-21 at 15 33 25

rodvar (approved):
ACK on all 3 apps

please have a look at my contrib commit buddha and let me know your thoughts @nostrbuddha

Issue #227, opened 2025-03-05
Offers: Clicking on a market with newly created offer sometimes won't show the offer


rodvar:
**Pre**: the market you are using has 0 offers. Let's suppose for example we have 0 offers with EUR. Also you need to have the offerbook filters sort by "Most offers"  & With offers.
Bug was reproduced using iOS Connect app but should happen with all of them

**Steps**:

1. open the app and go to the Offerbook tab
2. with another client (desktop for example), create an buy BTC offer with default settings in the EUR market
3. observe offerbook change, EUR is bumped up and has "1 offers" (minor b ug here, when offer qty is 1 it should be offer instead of offers)
4. tap on the EUR market and observe
5. go back, tap on each market till you find the offer

ER: EUR market should show the offer card to be able to take it
OR: Not only the offer is not there, but you will find it in other markets (say for example in the USD market) - see attached video for details in comments

rodvar:
https://github.com/user-attachments/assets/42301ac5-a72b-4540-8b8d-1c31b13609e9

HenrikJannsen:
Not looked into it, but often its that the filter predicate is not update or set on the changed collection.

rodvar:
fixed

PR #247, opened 2025-03-20
 - Fix for offer/market mismatch in xClients (#227)

nostrbuddha:
- Fix for https://github.com/bisq-network/bisq-mobile/issues/227
- `selectedMarket` was updated after offerlist is assigned in `applyOffersToSelectedMarket()` using the previously selectedMarket.


rodvar (approved):
ACK on all 3 apps

elegant right fix @nostrbuddha 💪

PR #254, opened 2025-03-25
Fix bug with calling connectivity service too early

HenrikJannsen:


rodvar:
rebased to main

rodvar (approved):
ACK

elegant fix thanks Henrik

Issue #136, opened 2024-12-20
Trusted Node Versioning Validation


rodvar:
For `androidNode` we manage versioning directly using gradle because we have static control of it.

For the `xClients`, after the first release we will need some sort of versioning handshake  to manage the app bisq required version (`R`) and the trusted node runing version (`V`):

 - R <= V. -> proceed normally
 - R > V -> Overlay error requesting the user to upgrade the trusted node, only option is to quit the client

From a UX perspective, I think the best is to present the user with the TrustedNode onboarding setup screen to give the option to connect to other trusted node, and also use the test button to verify it can connect and check if the trusted node upgraded its running version

rodvar:
@nostrbuddha would you like to finish this one, just missing the UI 😉

nostrbuddha:
Sure. Will do this

PR #261, opened 2025-03-28
check api compatibility with trusted node

rodvar:
- contrib for #136 
 - validation of api in selected trusted node
 - logs compatible/incompatible
 - all ready for UI implementation (from ClientMainPresenter)

### DIFF
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/bootstrap/ClientApplicationBootstrapFacade.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/bootstrap/ClientApplicationBootstrapFacade.kt
@@ -15,6 +15,10 @@ class ClientApplicationBootstrapFacade(
 
     private val backgroundScope = CoroutineScope(BackgroundDispatcher)
     override fun activate() {
+        if (isActive) {
+            return
+        }
+        log.d { "Running bootstrap." }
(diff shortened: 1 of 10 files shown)


rodvar:
## https://altstore.io/ option

### pros

 - approved by apple, but less restrictive listing approval process
 - Open-source & community-driven.
 - as easy as making an ipa and a descriptive json available and doing the signup
 - instalation for users has some friction if they are not subscribed to their patreon ($1/month at the time of writing)

### cons

 - free users will need to connect their iphone to their mac to revalidate signing once a week (generally tech savy users audience)
 
 ## Adding PWA (w/Tor support) as another UI access to bisq-mobile

### pros

 - opens up a wider range of users using browsers.
 - PWA is installable as an app typically with chrome-like and safari web browsers

### cons

 - would certainly need more dev time (test experimental Compose4Web, if it works no UI rewrite needed, otherwise UI screens rewritten in React + HTML)
 - most Tor browsers restrict usage of service workers (which are key to PWA), so we need to implement fallback support
 - other limitations apply for a typical PWA: IndexedDB, CDNs, google fonts, etc -> we'll need to implement our own self-hosted solutions
 - will need own hosting setup for deployment (this is also a pro in some ways)
 
 
 Other options I dismissed in this analysis because either need user to jailbreak, or it won't be a fit for us: sideloadly/trollstore, enterprise distrib, testflight public release, Cydia / Jailbreak store

Issue #55, opened 2024-11-15
Trading: Chat mediation


rodvar:
Private chat messaging support for MVP trade chat related only.

rodvar:
MVP is only trading-related chat so I'll modify the description to include this for the MVP (is the under the hood support for that private chat)

rodvar:
edited task to reflect what we are actually doing to finish the trading process

HenrikJannsen:
With https://github.com/bisq-network/bisq-mobile/pull/277 this is completed

PR #266, opened 2025-03-30
Add mediation support

HenrikJannsen:


rodvar:
implements #55

rodvar:
Hey @HenrikJannsen ! I did a first round of testing and found the following, let me know how would you like to handle them? (fix them in this PR, or lodge the issues to tackle later and merge as is since the main functionality is still usable)

## Connect
 
- not implemented, this PR implements mediation for node
- UI issues were found with certain devices (like iphone 13 mini), we could leave this for the connect implementation of mediation

## Node

 - The "Request mediation" button doesn't look like its clickable (following the same pattern as the rest of the app)
 - Request mediation button works great when there is a mediator avail. When there isn't, it let's you request but nothing happens. If you do the same on desktop it won't let you request it (we should do a synced impl)
 - after requesting mediation, trying to cancel the trade won't work anymore which is fine but no UI comm is given to the user (either disable or change warning text)

 thank you!
Screenshot 2025-03-31 at 13 39 57
Screenshot 2025-03-31 at 13 39 36

Screenshot 2025-03-31 at 13 45 43
Screenshot 2025-03-31 at 13 43 35
Screenshot 2025-03-31 at 13 40 45

HenrikJannsen:
I just tested cancelling after mediation requested and it worked.
Screenshot 2025-03-31 at 12 26 23

Handling of no mediator available could be done via a new issue, it should show a warning at take offer, and then hide the request button.

HenrikJannsen:
I moved both buttons into the expanded view as otherwise they would take too much space and attention. if anyone has a better UX solution please share. Not super happy that they are more hidden.

rodvar:
> I moved both buttons into the expanded view as otherwise they would take too much space and attention. if anyone has a better UX solution please share. Not super happy that they are more hidden.

I saw that as well, I think its ok since cancellation and mediation should be the exception and not the rule for a trading. 
What do you think @nostrbuddha  ?

rodvar:
@HenrikJannsen merging as is as per comments above

PR #274, opened 2025-04-02
 - Trusted node - UI for API version mismatch

nostrbuddha:
- UI for https://github.com/bisq-network/bisq-mobile/issues/136
 - On app start, popup is shown on APi version mismatch. Offers 'Fix' button, that navigates to 'Trusted node config' screen.
 - Trusted node config screen, now also checks for API version compatibility.
 - If connection happens, but API version is incompatible, respective error is shown and `next / save` button is hidden.

Question:
 - Right now, user can navigate to other screens, when API version mismatch. Should lock the user to Trusted node config screen, till connecting with a API compatible node?


rodvar:
rebased to main

rodvar:
@nostrbuddha answering your question in the description, yes that's the ideal behaviour. But let me test your implementation first we might have a good enough first version and then we can improve it later. Testing now.

rodvar (approved):
ACK

some nits

 - copy could be improved for UX (i'll change this now)
 - on small screen devices the container does not enlarge to fit all the text and user should scroll not leveranging the whole screen size space for a dialog. (we can fix this nit later)

PR #276, opened 2025-04-07
Remove not used topics and gateways

HenrikJannsen:


rodvar:
dropped?

HenrikJannsen:
Yes, have overlooked that parts are still used...

PR #283, opened 2025-04-11
Extract _genericErrorMessage out from MainPresenter

HenrikJannsen:


rodvar:
already merged in the follow-up PR #284

Issue #345, opened 2025-05-21
Supported languages from Settings are not considered while creating offer


nostrbuddha:
1. Go to Settings :: General settings.
2. Add couple of supported languages
3. Now create an offer
4. Offer card shows only `En` as supported languages.

Should take in the supported languages from Settings, while create the offer

rodvar:
@nostrbuddha you mean creating the offer is not considering settings supported languages, correct?

nostrbuddha:
Yes

PR #385, opened 2025-06-12
UI polish: Notification sync, Supported languages in Create offer

nostrbuddha:
1. [Fix] Notification: Be in My Trades tab. When someone takes my offer, the badge count increments immediately in bottom bar. But the trade card is not rendering. It renders only when switching between views.
 2. Create offer flow takes `General settings :: Supported languages` into account. (https://github.com/bisq-network/bisq-mobile/issues/345)

## Summary by CodeRabbit

- **New Features**
	- Offer creation now dynamically supports multiple languages based on user settings, rather than being limited to English.
- **Bug Fixes**
	- Corrected the displayed amount in the offer review screen to show the correct "to receive" value for sell offers.
- **Improvements**
	- Enhanced trade list updates to better reflect unread messages and open trades, and improved state management when navigating away from the trade list.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/di/PresentationModule.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/di/PresentationModule.kt
@@ -137,7 +137,7 @@ val presentationModule = module {
     factory { TakeOfferReviewPresenter(get(), get(), get()) }
 
     // Create offer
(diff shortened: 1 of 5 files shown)


rodvar (approved):
ACK all 3 apps

Visually the max amount of supported languages to be shown in a card is 4 which is acceptable. You can test this by configuring 5 supported languages in the settings and creating a new offer.

PR #394, opened 2025-06-29
Fix/UI fixes 01

nostrbuddha:
- Fix: BTC value in create offer review screen, didn't take premium percentage into account.
 - Fix: Fast double tap leads to double navigates with certain buttons in OpenTradeScreen.
 - Fix: TextField: Disabled state style - Improved to become more readable.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to expand or collapse trade details using a toggle button, with the state now managed by the presenter for consistency.
  * The floating chat button on the open trade screen is now enabled or disabled based on the interactive state of the screen.

* **Improvements**
  * The "Open wallet guide" button is now disabled when user interaction is not allowed.
  * Updated the appearance of disabled text fields with improved color contrast for better readability.

* **Bug Fixes**
  * Base-side coin values in offer creation are now dynamically calculated from fiat values and the current price quote, ensuring accurate conversions.


rodvar (approved):
ACK all 3 apps

PR #400, opened 2025-07-10
Start kmp-tor as external tor instance and provide config to bisq 2 tor lib.

HenrikJannsen:
Based on concept implemented in https://github.com/bisq-network/bisq-mobile/pull/398 but avoiding the extra complexity with a proxy server and redundant tor control message handling.

We start kmp-tor, apply the sock port, control port and auth file path to the configure file and start the network service.
The Bisq 2 tor lib will use that data for connecting to the external tor instance.

The android.conf  files have `devMode` set to false and `supportedTransportTypes` set to `TOR`.

We should provide some easy to manage way to switch between tor and clearnet. In Android Studio there is no option to set JVM arguments to override those config values. Maybe different run configs would be an option?

There are 2 open TODOs:
- Apply the state changes to the bootstap (we should IMO mirror the current bisq 2 bootstrap UI, which is a separate and more involved task) 
- Handle the error cases.

Btw. I could not get clearnet running with current main.

## Summary by CodeRabbit

* **New Features**
  * Integrated embedded Tor support, enabling secure network transport via Tor within the Android app.
  * Added a dedicated Tor service to manage the lifecycle and configuration of the embedded Tor runtime.
  * Introduced custom error handling for Tor operations.

* **Configuration**
  * Default network transport switched from "CLEAR" to "TOR".
  * Enabled use of an external Tor instance in configuration files.
  * Updated supported authorization features and token types.
  * Disabled developer mode in Android configuration.

* **Build and Dependency Updates**
  * Added dependencies for embedded Tor runtime and resources.
  * Limited supported architectures for native libraries and improved packaging to resolve resource conflicts.
  * Introduced build configuration fields for Tor version tracking and logging control.
  * Updated Java SDK version to 21.0.6.fx-zulu.

* **Bug Fixes**
  * Improved handling of native library conflicts during packaging.

### DIFF
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/KmpTorService.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/KmpTorService.kt
@@ -79,7 +79,10 @@ class KmpTorService(baseDir: Path) : Logging {
     }
 
     fun stopTor() {
-        require(torRuntime != null) { "torRuntime is null. setupTor must be called before stopTor" }
+        if (torRuntime == null) {
+            log.w("Tor runtime is already null, skipping stop")
+            return
+        }
 
         torRuntime!!.enqueue(
             Action.StopDaemon,
@@ -199,21 +202,35 @@ class KmpTorService(baseDir: Path) : Logging {
 
                 var lastModified = 0L
                 var iterations = 0
-                val maxIterations = 100
+                val delay: Long = 100
+                val maxIterations = 30 * 1000 / delay // 30 seconds with 100ms delays
+                val startTime = System.currentTimeMillis()
+                val timeoutMs = 60_000 // 60 second timeout
                 while (!deferred.isCompleted && iterations < maxIterations) {
                     iterations++
+                    if (System.currentTimeMillis() - startTime > timeoutMs) {
+                        deferred.completeExceptionally(
+                            KmpTorException("Timeout waiting for control port file")
+                        )
+                        break
+                    }
                     log.i("readControlPort iterations=$iterations")
                     if (controlPortFile.exists()) {
                         val currentModified = controlPortFile.lastModified()
                         if (currentModified != lastModified) {
                             lastModified = currentModified
-                            log.i("We detected modification change of controlPortFile: $controlPortFile ; iterations=$iterations")
+                            log.i("We detected modification change of controlPortFile: $controlPortFile")
                             readControlPortFile(controlPortFile, deferred)
                         }
                     }
                     if (!deferred.isCompleted) {
-                        delay(100)
+                        delay(delay)
                     }
+                }
+                if (!deferred.isCompleted) {
+                    deferred.completeExceptionally(
+                        KmpTorException("Failed to read control port after $iterations iterations")
+                    )
                 }
             } catch (e: Exception) {
                 handleError("Observing file ${controlPortFile.absolutePath} failed: ${e.message}")


rodvar:
Instead of working towards merging this, I'm gonna review & test the enhancements and bring the enhancements that work in my PR (referenced in the description) since that PR has everything already working - hope this is ok 🙏 

We can then test the integrated solution in that PR and merge it in 🚀

rodvar:
rebased to main

rodvar:
Answering comments in description:

`We should provide some easy to manage way to switch between tor and clearnet. In Android Studio there is no option to set JVM arguments to override those config values. Maybe different run configs would be an option?`

 we currently have 2 android.conf, one used for release builds and the main one use for any other (typically debug). That's why the main android.conf should be setup for CLEARNET as default, leaving to the dev to replace it if he wants to test a debug build against prod.
 
 Eventually we can add a 3rd config (called "staging" ? ) so that we don't need to manually modify the file 👍 
 
 `Apply the state changes to the bootstap (we should IMO mirror the current bisq 2 bootstrap UI, which is a separate and more involved task)` -> there is an issue created in the backlog for this, probably will be done in the milestone immediately after the current release one.
 
 `Handle the error cases.` -> I will take care of this ;)

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/presentation/NodeMainPresenter.kt
```diff
@@ -81,6 +90,15 @@ class NodeMainPresenter(
                     provider.applicationService = applicationService
 
                     applicationBootstrapFacade.activate()
+
+                    // todo apply kmp tor startup state to applicationBootstrapFacade
```
rodvar, line 94: can do this in my branch after merging

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/KmpTorService.kt
```diff
@@ -0,0 +1,319 @@
...
+ * 2. Start the kmp tor runtime
+ * 3. Find socksPort by listening on TorEvent.NOTICE with data: `Socks listener listening on port {socksPort}.`
+ * 4. Find control port by observing creation or mutation of the `control.txt` file which contains the control port.
+ *    Read the value from format: `PORT=127.0.0.1:{controlPort}`
+ * 5. Write the `external_tor.config` with socks port, control port and path to the auth cookie. This will be used by the Bisq 2 tor lib.
+ * 6. After tor daemon is started, we are completed. The network service can now be initialized.
+ *    The bisq 2 tor lib will detect the external tor and use that.
+ *
+ */
+class KmpTorService(baseDir: Path) : Logging {
```
rodvar, line 41: should extend our BaseService and leverage coroutine helpers, can refactor in my branch

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/KmpTorService.kt
```diff
@@ -0,0 +1,319 @@
...
+            deferred.completeExceptionally(error)
+        }
+    }
+
+    private fun writeExternalTorConfig(socksPort: Int, controlPort: Int) {
+        try {
+            val cookieFile = File(getTorDir(), "control_auth_cookie")
+            val configContent = buildString {
+                appendLine("UseExternalTor 1")
+                appendLine("CookieAuthentication 1")
```
rodvar, line 265: great

androidNode/src/main/resources/android.conf
```diff
@@ -116,8 +116,8 @@ application {
         # FOR DEV: uncomment the one you need to test.
         # Bear in mind TOR will use prod network, if you have a profile already installed
         # you need to wipe the app storage to trigger the profile creation on the next start
-        supportedTransportTypes = ["CLEAR"]
-        features = ["INVENTORY_HASH_SET","AUTHORIZATION_HASH_CASH"]
+        supportedTransportTypes = ["TOR"]
+        features = ["INVENTORY_HASH_SET", "AUTHORIZATION_HASH_CASH", "AUTHORIZATION_HASH_CASH_V2"]
```
rodvar, line 120: replaced lines are preferred since this main android conf is used for dev so this will be reverted in my branch
later on we can implement multiple flavours with different android.conf setups if its worth it

rodvar (approved):
ACK

I've changed my mind after reviewing and testing this PR. As the description says its a curated approach that brings the best of the concept PR removing the complexities I got into whilst discovering the solution.

This PR breaks release builds but that's ok cause I have it fixed in my branch. 
It will bring lots of conflicts to my branch I think but I'll deal with that, creating a new branch if necessary.

Thanks Henrik, great work!

PR #449, opened 2025-07-17
fix(UI): remove logo & adjust avatar size

### LINKED ISSUE #442: Create Profile: UI adjustments
Image

 - Logo is not needed at the stage where the user is creating the profile
 - make profile image larger.
 - min lenght for the nickname is not really required (it can be empty).

wodoro:
and allow nicknames with at least 1 chars
in profile creation screen

fixes #442


wodoro:
image

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/startup/CreateProfilePresenter.kt
```diff
@@ -79,7 +79,7 @@ open class CreateProfilePresenter(
 
     fun validateNickname(nickname: String): String? {
         return when {
-            nickname.length < 3 -> "Min length: 3 characters"
+            nickname.length < 1 -> "Min length: 1 characters"
```
rodvar, line 82: nit - character instead of characters

PR #459, opened 2025-07-24
fix: search bar clear & btn weights

### LINKED ISSUE #423: Offerbook: Search "x" showing up even if there is no input entered
Image

as reported by Henrik: `as clear (x) and filter is close together and x only needed when input in search is added we should display x only when input it there and hide filter in that case. otherwise is only filter`

### LINKED ISSUE #424: Open Trade Chat: minor ui adjustments
Image
Image

 - reduce size of avatars to match desktop style
 - center input and justify to container

wodoro:
fix #423
the button were actually not in equal size and the filter button would move unnecessarily without the spacer when weights are adjusted, so I thought it would be better to fix those issues as well by aligning the content to the right and setting the weight modifiers equally

here's a video comparison:
old:

[old.webm](https://github.com/user-attachments/assets/3bced3ca-cf06-4a4d-b374-10f26ed310f8)

new:

[new.webm](https://github.com/user-attachments/assets/3eaa8770-de3a-4c81-8a38-4a18306be5af)

## Summary by CodeRabbit

* **Refactor**
  * Improved the layout stability of search and text fields, ensuring consistent alignment and spacing for suffix elements and clear buttons.
* **Style**
  * Enhanced alignment options for text field components, allowing for more precise control over suffix content positioning.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/components/atoms/TextField.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/components/atoms/TextField.kt
@@ -52,6 +52,7 @@ fun BisqTextField(
     leftSuffix: (@Composable () -> Unit)? = null,
     rightSuffix: (@Composable () -> Unit)? = null,
     rightSuffixModifier: Modifier = Modifier.width(50.dp),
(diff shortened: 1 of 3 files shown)


rodvar (approved):
ACK all 3 apps

PR #466, opened 2025-07-27
fix: show close button in offer creation wizard

### LINKED ISSUE #406: Create Offer: Navigation enhancements
Image

Notice the current create offer workflow:

 - There is no way to fully exit the workflow, if the user its on its final step and decides to cancel everything, he needs to go back several times or kill the app

Proposals for fixing this:

 1. hide the avatar whilst in the workflow and replace with our closing "x" that will prompt the user with a confirmation dialog to cancel the offer creation when tapped OR
 2. since we already handle back navigation with the common mobile "<" arrow pattern, instead of having also the back at the bottom in the main control menu, replace the back button with a "cancel" with the same tap behaviour

The goal of this issue is to decide which approach and implement it.

wodoro:
resolves #406

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/components/atoms/button/IconButton.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/components/atoms/button/IconButton.kt
@@ -13,13 +13,12 @@ import network.bisq.mobile.presentation.ui.theme.BisqUIConstants
 @Composable
 fun BisqIconButton(
     onClick: () -> Unit,
-    modifier: Modifier = Modifier,
+    modifier: Modifier = Modifier.size(BisqUIConstants.ScreenPadding2X),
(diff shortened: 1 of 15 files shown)


rodvar (changes requested):
Well implemented improvement.

It only needs one minor fix to be merged. Is not a fix introduced in this PR, rather using this PR to fix a longstanding UI bug:

Screenshot 2025-07-28 at 10 16 10
Screenshot 2025-07-28 at 10 13 17


both avatar and now close buttons are "falling down" a bit. Thew should be vertically centered as the back arrow is

wodoro:
> both avatar and now close buttons are "falling down" a bit. Thew should be vertically centered as the back arrow is

Actually I could fix that, but it seemed intentional to me (because of paddings added), that's why I did not touch 

so you think I should center it ?

wodoro:
image
image

this is with the intentional top padding removed

rodvar (approved):
ACK

PR #474, opened 2025-07-29
feat: add chat rules screen

wodoro:
will be accessible from chat warning box, and later from inside report user screen:

image

image


rodvar:
rebased to main

rodvar:
> Actually, I think you are better suited as the change seems significant and I don't have any idea right away, can you look into it @rodvar ?

thanks so much for having a look into this, yep that's the commit I suspect of particularly the change on the Topic related to chat might be incomple - don't worry I'll find the fix for it today 🙏

PR #492, opened 2025-08-01
fix: change TextField label color

### LINKED ISSUE #421: Open Trade Flow: lookup data info hardly readable
Image

it's because of a combination of background color, color and font that the "lookup..." messages are hardly readable

A small tweak on color might just do the trick to improve it.

### LINKED ISSUE #491: User Profile: headers are hardly visible
Image

see screenshot, headers need more contrast as it was done for the trade flow screens

wodoro:
I simply changed the default label color across the app to be brighter in general and it looks like it solves the problem
resolves #421
resolves #491

image

image

## Summary by CodeRabbit

* **Style**
  * Updated the color of the label in disabled text fields for improved visual appearance.


rodvar:
rebased to main

rodvar (approved):
ACK - neat fix with default color change on our UI components

PR #509, opened 2025-08-07
feat: add notification icon

### LINKED ISSUE #497: Background notifications: add monochrome logo
currently background notifications are showin with a default white ring
we need to add the ic_notifications.xml svg bisq monochrome logo and add it in the generated notifications for android

wodoro:
resolves #497 

It seems I was wrong
I tried removing the explicit icon but it didn't use the correct icon (ic_launcher did not work), and even if it did it would depend on android version,
so I think the only way was to actually add an icon for it, so I did

this was the only way that came to my mind for implementing it
feel free to suggest a better way if you know any


rodvar:
rebased to main

rodvar (approved):
ACK all android apps

answering your comment in the description: this is exactly how I would have implemented it 👍

PR #547, opened 2025-08-15
Improve ws connection setup

HenrikJannsen:
Based on https://github.com/bisq-network/bisq-mobile/pull/543

Screenshot 2025-08-15 at 20 46 18

## Summary by CodeRabbit

* **New Features**
  * New WebSocket setup with separate Host/Port inputs and LAN/Tor selection.
  * Built-in “Test Connection” with clear status and semantic API version checks.

* **UI/UX**
  * Rebranded app surface to “Bisq Connect”; updated wording to “remote node”.
  * Overhauled onboarding with Full Mode/Client Mode content.
  * New WebSocket setup screen with improved prompts and validation.
  * Updated node dashboard texts; progress indicator styling refined.

* **Bug Fixes**
  * Corrected startup flow to prompt profile creation only when no profile exists.

### DIFF
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/settings/ClientSettingsServiceFacade.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/client/service/settings/ClientSettingsServiceFacade.kt
@@ -136,6 +136,7 @@ class ClientSettingsServiceFacade(private val apiGateway: SettingsApiGateway) : 
         val trustedNodeApiVersion = apiGateway.getApiVersion().getOrThrow().version
         log.d { "required trusted node api version is $appApiRequiredVersion and current is $trustedNodeApiVersion" }
         // return false // (for debug)
+        //todo lexical order nto safe here
         return trustedNodeApiVersion >= appApiRequiredVersion
     }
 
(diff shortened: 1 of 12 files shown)


rodvar:
> > Visual changes and wording are an improvement, but this PR can be merged as is due to the following issues:
> > ```
> > * Tor/Lan toggle needs to be disabled/hidden/commented until we actually support tor in connect
> > ```
> 
> I think @wodoro is working on that and I expect that this will be ready for release time. But if you like we can hide it. But I prefer to do that in a follow up PR, as its painful to maintain that chain of PRs....
> 
> > ```
> > * back navigation broken: back arrow should be hidden when the screen is presented in workflow, and shown when accessing from settings
> > ```
> 
> Accessing it from Settings does not work yet, So I kind of ignored that case. I think its better to copy paste the screen and adjust to whats needed when used from settings as the current screen is a mix of onboarding and setting of the address. I think its less pain to have a bit duplication than to support 2 different use cases.
> 
> > ```
> > * after successful connection, if the user decides to change to a different trusted node state of UI doesn't refresh and create profile button is still shown ,progressing uses the previous node setup
> > ```
> 
> Is this really a realistic use case? Once you get connected you are happy and want to move on. But if wanted I would also prefer to do it in a follow up, to avoid another merge/rebase hell cycle.
> 
> > I can try fix all these and merge it but probably better to do at the end of the chain of PRs
> 
> Yes, please dont add any commit as otherwise I need to update all the others as well...

If everything goes well as its going we are looking into releasing node this month (at least a soft launch, to main Bisq2 traders and then release publicly from there). When that happens will bring attention back to connect apps and do the same release process so its likely there will be enough time to incorporate tor 👍 . But in the meantime I'd like to keep the main branch showing only the features that are ready for release.

` Is this really a realistic use case?` -> what we do on main is automatically proceed to the profile screen after successful connection to avoid this situation.

I'll figure out how to do the merge in the last node of the PR chain 👍

rodvar:
closing in favour of https://github.com/bisq-network/bisq-mobile/pull/549

Let's please try to avoid chained PRs , prefer isolated changes based on main integration branch 🙏

PR #630, opened 2025-09-08
Add transifex

hiciefte:
This adds the Transifex configuration for the mobile specific translations and also converts the translation files to UTF-8. This is setup is required, so the mobile specific project translations can be synced and translated automatically.


hiciefte:
Obviously a couple of errors slipped through, I'll fix them asap

rodvar:
Testing the app before merging and the bootstrap gets stucked in "Bootstrap to P2P network", I'm investigating now

Issue #707, opened 2025-09-17
Warn user if bitcoin explorer transaction amount is not matching


rodvar:
> It is also missing that overlay if amount was not matching after clicking next.
> <img width="1650" height="800" alt="Image" src="https://github.com/user-attachments/assets/1c0e726d-005e-4135-92b4-0c5ed87a3e53" /> 

 _Originally posted by @HenrikJannsen in [#706](https://github.com/bisq-network/bisq-mobile/issues/706#issuecomment-3301899419)_

missing small feature, warning as per ☝️

rodvar:
@salottodev will be tackling this one tomorrow

PR #714, opened 2025-09-18
Warn on transaction amount mismatch and refactor presenters

salottodev:
Fixes: https://github.com/bisq-network/bisq-mobile/issues/707

This PR adds a warning dialog when the detected transaction amount does not match the expected trade amount.

As part of this change, the Buyer and Seller presenters was extracted into a new BaseTradeStateMainChain3bPresenter. This eliminates code duplication and creates a single source of truth, improving long-term maintainability.

**Buying BTC**

https://github.com/user-attachments/assets/2014c78a-cc24-4a06-8f85-eaad9f22c41e

**Selling BTC**

https://github.com/user-attachments/assets/cdd9d5dc-ffc7-415a-82ce-66a0114af474


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/helpers/Utils.kt
```diff
@@ -3,6 +3,8 @@ package network.bisq.mobile.presentation.ui.helpers
 import okio.ByteString.Companion.encodeUtf8
 import kotlin.math.abs
 
+const val EMPTY_STRING = ""
```
rodvar, line 6: nice

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BaseTradeStateMainChain3bPresenter.kt
```diff
@@ -0,0 +1,224 @@
...
+import network.bisq.mobile.i18n.i18n
+import network.bisq.mobile.presentation.BasePresenter
+import network.bisq.mobile.presentation.MainPresenter
+import network.bisq.mobile.presentation.ui.helpers.EMPTY_STRING
+import network.bisq.mobile.presentation.ui.uicases.open_trades.selected.states.TxConfirmationState.CONFIRMED
+import network.bisq.mobile.presentation.ui.uicases.open_trades.selected.states.TxConfirmationState.IDLE
+import network.bisq.mobile.presentation.ui.uicases.open_trades.selected.states.TxConfirmationState.IN_MEMPOOL
+import network.bisq.mobile.presentation.ui.uicases.open_trades.selected.states.TxConfirmationState.REQUEST_STARTED
+
+// TODO: Add btc address/tx validation missing
```
rodvar, line 24: 👍  for node we can use the Bisq2 utils to do it
for connect we'll have to implement it..
Thinking long term, It would be cool if we could have a common jar to use in shared with all the bitcoin related utils at least, something to discuss with @HenrikJannsen 👍

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BaseTradeStateMainChain3bPresenter.kt
```diff
@@ -0,0 +1,224 @@
...
+abstract class BaseTradeStateMainChain3bPresenter(
```
rodvar, line 25: thanks for doing this refactor 🙏 💪 
lots of duplicated code in the create/take offer flows

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BaseTradeStateMainChain3bPresenter.kt
```diff
@@ -0,0 +1,224 @@
...
+    override fun onViewUnattaching() {
+        _buttonText.value = EMPTY_STRING
+        _txConfirmationState.value = IDLE
+        _blockExplorer.value = EMPTY_STRING
+        _balanceFromTx.value = EMPTY_STRING
+        _errorMessage.value = null
+        super.onViewUnattaching()
+    }
+
+    fun onCtaClick() {
```
rodvar, line 89: what does Cta stands for? Call to action? 
please use better naming related to the action or add header doc please

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BaseTradeStateMainChain3bPresenter.kt
```diff
@@ -0,0 +1,224 @@
...
+        return if (address != null && txId != null && txAmountFormatted != null) {
+            "bisqEasy.tradeState.info.phase3b.button.next.amountNotMatching".i18n(
+                address,
+                txId,
+                txAmountFormatted,
+                tradeAmountFormatted
+            )
+        } else {
+            null
+        }
```
rodvar, line 114: could be simplified using kotlin `takeIf` but actually this is more readable 👍

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BuyerStateMainChain3bPresenter.kt
```diff
@@ -1,167 +1,17 @@
...
-        }
-    }
+    tradesServiceFacade: TradesServiceFacade,
+    explorerServiceFacade: ExplorerServiceFacade
+) : BaseTradeStateMainChain3bPresenter(
+    mainPresenter,
+    tradesServiceFacade,
+    explorerServiceFacade
+) {
+    // This class is now a placeholder for any future Buyer-specific logic.
```
rodvar, line 16: for this particular case, the placeholders could be removed and the base class made a concrete class

rodvar (approved):
ACK

some nits marked on the code, if agreed can be done on a follow up PR

💪

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BaseTradeStateMainChain3bPresenter.kt
salottodev: That comment was not added by me, but I decided to keep it.

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/open_trades/selected/states/BaseTradeStateMainChain3bPresenter.kt
salottodev: Yes, it means call to action. I decided that name because the button is Genuinely a "Call to Action": The primary button on this screen is dynamic. Its text and purpose change based on the current state:
•Sometimes it says "Next".
•Sometimes it says "Skip". 
The term "Call to Action" (CTA) is a standard UX term for the main button that prompts a user to do something. Since this button serves multiple purposes, a generic name reflects its generic role in the UI.

But I agree with you that it could be confusing. May be a better naming would be: `onAdvanceStepClick()`

HenrikJannsen: If there is only one button for such we could just name it `onButtonClick`, the `onCtaClick` would also only work for one generic button... I personally am not a big fan of the acronyms as its less commonly knows to non-english native speakers where its less usual to use that many acronyms in the language as in English.

HenrikJannsen:
The text with the address and tx ID is a bit hard to read and long. I think specially for mobile with more restricted space and a potentially less technical orientated audience, we might consider to just drop those detail data as they are anyway present in the screen. We should likely apply that also on desktop.

Issue #604, opened 2025-08-30
Notifications: Open trades "New messages from your peer" notification and update notification badge


rodvar:
extracted from #600 

The apps currently let you know if there is a state progress in any open trades (or one of your offers gets taken)
This issue is to bring also notifications when there is new messages from a peer in one of your open trades.
Recent changes on chat might make the trigger to this easier as the app is currently identifying when there are new unread messages.

Also make sure that the tab badge displaying number of open offers with updates also considers this

rodvar:
@wodoro has started work on this to see if we can include it in the RC-6 pre release

A small nuance to mimick Bisq2 desktop:

Right now: we based our "My Trades" badge on "total number of open trades with unread notifcations"
Bisq2: its based on total number of unread events.

So in the picture above, instead of showing 1 (as in one open trade with unread notif) we should show 4 (because there are 4 new events unread) . If that trade would also have an unread state progression ("peer pay fiat") it would be 5

PR #720, opened 2025-09-19
feat: make all events count toward badge

wodoro:
this will make "My Trades" badge count up to the unread events
however the final state condition is causing different behaviors, I need to know why it was added, as comment in the code is not explaining it
so I will make this PR with old behavior (different from desktop) for final state
then make an additional commit that will make it aligned with desktop behavior

toward resolving #604

## Summary by CodeRabbit

- New Features
  - Bottom navigation badge now shows the total number of unread chat messages across all trades.
  - Profile avatars and trade list items display per-trade numeric unread badges.

- Bug Fixes
  - Unread counts exclude ignored senders and only count visible message types.
  - Badge totals are computed as the sum of unread messages and update consistently in real time.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/MainPresenter.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/MainPresenter.kt
@@ -66,8 +66,7 @@ open class MainPresenter(
                         trade.bisqEasyOpenTradeChannelModel.chatMessages,
                         tradeReadStateRepository.data.map { it.map },
                         userProfileServiceFacade.ignoredUserIds,
-                        trade.bisqEasyTradeModel.tradeState,
-                    ) { messages, readStates, ignoredUserIds, tradeState ->
+                    ) { messages, readStates, ignoredUserIds ->
                         // TODO: refactor to filter visible messages based on ignore at trade.bisqEasyOpenTradeChannelModel.chatMessages for consistency
                         // this is a duplicated logic from OpenTradePresenter msgCount collection
                         val visibleMessages = messages.filter {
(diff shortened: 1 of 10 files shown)


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/MainPresenter.kt
```diff
@@ -59,35 +61,35 @@ open class MainPresenter(
...
-                        .map { messages -> trade.tradeId to messages.size }
+            .flatMapLatest { openTradeItems ->
+                val flowsList = openTradeItems.map { trade ->
+                    combine(
+                        trade.bisqEasyOpenTradeChannelModel.chatMessages,
+                        tradeReadStateRepository.data.map { it.map },
+                        userProfileServiceFacade.ignoredUserIds,
+                    ) { messages, readStates, ignoredUserIds ->
+                        // TODO: refactor to filter visible messages based on ignore at trade.bisqEasyOpenTradeChannelModel.chatMessages for consistency
+                        // this is a duplicated logic from OpenTradePresenter msgCount collection
```
rodvar, line 74: 👍

rodvar:
ahh it makes sense from the variable changes, I'll push the fix now

rodvar (approved):
ACK (node)

this works beautifully as well. IMHO this closes #604 but I'm confused with your description comment, I'll wait for your feedback to close it

The enhancement on the colors is nice, a bit less harsh on the eyes and still clearly shows which offers have updates 👍 

Only nit - the only way to clearing the badges is going to the chat and see all the notifications. We might want to add some kind of way to clean them up from the trade details screen but not a blocker.

PR #757, opened 2025-09-26
bugfix/ui_review_6

### LINKED ISSUE #747: Take offer happens double time on fast click
In Take offer :: Review screen, fast tap on 'Confirm take offer', the offer is taken twice.

Of the two taken trades, one works good. And the other throws this error:

Image

### LINKED ISSUE #755: Trim nickname in Create your profile
currently if you accidentally leave a trailing space in profile nickname, it will prevent you from creating a profile with a error dialog

the better UX would be just trim it and continue

rodvar:
- fixes #755 
 - fixes #747 
 - always work with trimmed value for create profile nickname validations and broadcasting
 - tackled corner case in between take offer dialogs allowing for a double take with a shot interactivity delay, applied to both confirmations screens

## Summary by CodeRabbit

- Bug Fixes
  - Briefly block UI interactivity on review screens to prevent accidental double submissions (take/create offer flows).
  - Trim leading/trailing spaces from nicknames during entry, validation, and submission to avoid invalid profiles.

- Tests
  - Added a unit test verifying interactivity is briefly blocked on view attach and then re-enabled.

### DIFF
--- a/shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/ui/uicases/take_offer/TakeOfferDoubleSubmissionTest.kt
+++ b/shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/ui/uicases/take_offer/TakeOfferDoubleSubmissionTest.kt
@@ -1,76 +1,86 @@
 package network.bisq.mobile.presentation.ui.uicases.take_offer
 
-import io.mockk.coEvery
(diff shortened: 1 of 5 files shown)


rodvar:
> just an idea, maybe it would be better if we block the interactivity globally ?

happy to explore this in the future thanks wod, if you wanna picture what you are thinking on this one ping me -  for now presentation level one works fine

PR #764, opened 2025-09-27
Remove use-animations from settings

HenrikJannsen:
## Summary by CodeRabbit

- Chores
  - Simplified Settings by removing the Display/Animations section, including its header, divider, spacing, and the “Use animations” switch. The animations preference can no longer be changed from within the app, and the toggle is no longer visible. Other settings remain unaffected, with no additional functional changes beyond this UI cleanup.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/settings/SettingsScreen.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/settings/SettingsScreen.kt
@@ -167,18 +167,6 @@ fun SettingsScreen() {
             }
         )
 
-        BisqHDivider()
-
-        BisqText.h4Light("settings.display.headline".i18n())
-
-        BisqGap.V1()
-
-        BisqSwitch(
-            label = "settings.display.useAnimations".i18n(),
-            checked = useAnimations,
-            onSwitch = { presenter.setUseAnimations(it) }
-        )
-
         if (shouldShowPoWAdjustmentFactor) {
             BisqHDivider()
 


rodvar:
nACK

I respectfully disagree with this change for the following reasons:

 - Animations are on by default so its good to give the user the possibility to change them
 - While its true that for node we could go without it (target hi end devices), connect apps can run in low end devices so for the same reason we have this toggle in Bisq2 desktop
 - This is a working functionality that's been here since the start of the project, if after getting feedback from users we found it doesn't make sense I would agree but removing it based on personal preference is not a good enough reason IMHO - we build this product for people with different preferences and backgrounds.

HenrikJannsen:
Ok, no that important to waste too much energy on that... You can close it.

Just to re-iterate my central point. 
Every options comes with mental costs. The user need to recognize the feature, understand (and that is specially problematic here as even I did not understand why when I turn it off I still see animations, I considered it a bug) and remember it. The more you add such costs the more difficult the app will feel. A feature need to provide more value as the costs.

Flooding users with options is not good UX. See Linux versus Apple. If our target audience is only the tech geeks, that can be ok, as they have different preferences, but with Bisq Easy and Mobile we clearly try to reach a more non-technical audience.

rodvar:
> Ok, no that important to waste too much energy on that... You can close it.
> 
> Just to re-iterate my central point. Every options comes with mental costs. The user need to recognize the feature, understand (and that is specially problematic here as even I did not understand why when I turn it off I still see animations, I considered it a bug) and remember it. The more you add such costs the more difficult the app will feel. A feature need to provide more value as the costs.
> 
> Flooding users with options is not good UX. See Linux versus Apple. If our target audience is only the tech geeks, that can be ok, as they have different preferences, but with Bisq Easy and Mobile we clearly try to reach a more non-technical audience.

Thanks for understanding and for explaining the reasoning behind it.

I certainly agree with not flooding user with settings but I don't think is a big deal in this case. Also the reaction is immediate as they can see the avatar stop shinning when turning off and starting when turning on.

If we find users complaining about this in any form I'll be the first one to propose removing it 👍 

But honestly I don't think they will care about this, it would be more about bug reports & feature/enhancement requests (hopefully for of the latter :)

PR #776, opened 2025-09-30
Add MemoryProfiler

HenrikJannsen:
Writes in debug mode a file to the data dir (memory_log.csv), which can be opened in a spreadsheet app and applied to a chart to see the memory footprint over time.

As Android profiler is a pain to use, that might be an easier and low-overhead way to get insight about memory consumption and help in optimisations.

If not wanted, feel free to close the PR. Just wanted to share it if others find it useful. I can keep it as local branch and use it when needed...


rodvar:
Thanks for sharing Henrik!

I personally feel comfortable with Android Studio profiler. But somebody else in the team might like this so I'll leave it in the post release improvements milestone and will give it a review later on 🙏

rodvar:
Had a chat with the team and nobody has intention to use it (either don't do profiling or use Android Studio like me) so I'm dropping it now - thanks for sharing though

PR #777, opened 2025-09-30
Improve offers service facade

HenrikJannsen:
This will require latest Bisq 2 mobile branch to be published.

Some aspects handled here are discussed at https://github.com/bisq-network/bisq-mobile/discussions/780.

From memory monitoring I could not see any significant difference, but there was some spike at initialization. 
By not having any channel selected initially we can avoid that. Only when the user browse to offerbook and select a market the list gets filled. We do not set the channel to null when the user move away from the offerbook, as it seems not needed. The number of offers is low (about 50 for USD) and I could not see any signature in the memory logging when selecting a market.

The extra handling with a queue and delays is not needed as it is not much data, thus this was removed to reduce complexity.

There is though a memory spike at the very start of the application, caused likely we we load all persisted data. See https://github.com/bisq-network/bisq-mobile/issues/807.
 

## Summary by CodeRabbit

- Refactor
  - Removed mediator-wait pathway and dialog; offer creation now uses a single streamlined flow with standard error notifications.
  - App startup initialization simplified for more consistent startup behavior.
  - Offer list/count updates moved to event-driven, per-message updates for faster UI refreshes and reduced batching.

- New Features
  - Chat/message validation plus ignore/ban checks applied to offers and market counts.

- Chores
  - Removed obsolete mediator-wait localization string.

### DIFF
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
@@ -556,7 +556,7 @@ class NodeOffersServiceFacade(
                 try {
                     val offerId = message.bisqEasyOffer.get().id
 
-                    if (offerMessagesContainsKey(offerId) || !isValidOfferMessage(message)) {
+                    if (!isValidOfferMessage(message)) {
                         continue
                     }
 
@@ -591,14 +591,17 @@ class NodeOffersServiceFacade(
 
         // Single UI update for all new offers
         if (newOffers.isNotEmpty()) {
-            _offerbookListItems.update { it + newOffers }
+            // distinct avoids that we add duplicates but requires proper equals/hashCode in OfferItemPresentationModel
+            _offerbookListItems.update { currentList ->
+                (currentList + newOffers).distinct()
+            }
+
             val currentSize = _offerbookListItems.value.size
             log.i { "Batch processed $processedCount offers, total: $currentSize" }
 
             // Log memory pressure if list is getting large
             if (currentSize > 100 && currentSize % 50 == 0) {
-                val mapSize = offerMapMutex.withLock { offerIds.size }
-                log.w { "MEMORY: Large offer list - UI: $currentSize, Map: $mapSize" }
+                log.w { "MEMORY: Large offer list - UI: $currentSize" }
             }
         }
     }
@@ -627,11 +630,10 @@ class NodeOffersServiceFacade(
                     val runtime = Runtime.getRuntime()
                     val usedMemory = (runtime.totalMemory() - runtime.freeMemory()) / 1024 / 1024
                     val maxMemory = runtime.maxMemory() / 1024 / 1024
-                    val offerMapSize = offerMapMutex.withLock { offerIds.size }
                     val offersListSize = _offerbookListItems.value.size
                     val observersCount = numOffersObservers.size
 
-                    log.w { "MEMORY: Used ${usedMemory}MB/${maxMemory}MB, OfferMap: $offerMapSize, OffersList: $offersListSize, Observers: $observersCount" }
+                    log.w { "MEMORY: Used ${usedMemory}MB/${maxMemory}MB, OffersList: $offersListSize, Observers: $observersCount" }
 
                     // Only suggest GC in critical situations (90%+) to avoid masking memory leaks
                     if (usedMemory > maxMemory * MEMORY_GC_THRESHOLD) {
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
@@ -61,6 +61,7 @@ class NodeOffersServiceFacade(
     companion object {
         private const val SMALL_DELAY = 25L
         private const val SMALL_DELAY_THRESHOLD = 5
+
         // Higher threshold to avoid masking memory leaks - only suggest GC in critical situations
         private const val MEMORY_GC_THRESHOLD = 0.85
         private const val OFFER_BATCH_DELAY = 100L // milliseconds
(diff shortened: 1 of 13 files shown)


androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -277,163 +206,111 @@ class NodeOffersServiceFacade(
...
+         public boolean isValid(BisqEasyOfferbookMessage message) {
+        return isNotBanned(message) &&
+                isNotIgnored(message) &&
+                (isTextMessage(message) ||
+                        isBuyOffer(message) ||
+                        hasSellerSufficientReputation(message));
+    }
+         */
+        return true
     }
```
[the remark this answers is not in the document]
rodvar: I think I can implement this fairly quickly, I'll do it in my pre-merge commit

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -291,19 +327,63 @@ class NodeOffersServiceFacade(
...
-                        isBuyOffer(message) ||
-                        hasSellerSufficientReputation(message));
+            isNotIgnored(message) &&
+            (isTextMessage(message) || isBuyOffer(message) || hasSellerSufficientReputation(message))
+    }
+
+    private fun isNotBanned(message: BisqEasyOfferbookMessage): Boolean {
+        val authorUserProfileId = message.authorUserProfileId
+        val profileOpt = userProfileService.findUserProfile(authorUserProfileId)
+        if (!profileOpt.isPresent) {
```
HenrikJannsen, line 339: Better use `profileOpt.isEmpty`

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -291,19 +327,63 @@ class NodeOffersServiceFacade(
...
+            log.w { "isNotBanned: author profile not found for id=$authorUserProfileId" }
+            return false
```
HenrikJannsen, line 341: I think its better to return true here as we dont know if he is banned. But it would not make any real difference as we dont show any messages/offer of profile is not found.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -291,19 +327,63 @@ class NodeOffersServiceFacade(
...
         return isNotBanned(message) &&
-                isNotIgnored(message) &&
-                (isTextMessage(message) ||
-                        isBuyOffer(message) ||
-                        hasSellerSufficientReputation(message));
+            isNotIgnored(message) &&
+            (isTextMessage(message) || isBuyOffer(message) || hasSellerSufficientReputation(message))
+    }
+
+    private fun isNotBanned(message: BisqEasyOfferbookMessage): Boolean {
```
HenrikJannsen, line 336: As the facade is just mirroring the service I think its better to stick with the service method names, like `isUserProfileBanned` in that case. Applies to the other methods as well.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -291,19 +327,63 @@ class NodeOffersServiceFacade(
...
+        val authorUserProfileId = message.authorUserProfileId
+        val profileOpt = userProfileService.findUserProfile(authorUserProfileId)
+        if (!profileOpt.isPresent) {
+            log.w { "isNotBanned: author profile not found for id=$authorUserProfileId" }
+            return false
+        }
+        return !bannedUserService.isUserProfileBanned(profileOpt.get())
```
HenrikJannsen, line 343: We have that method in the service: `isUserProfileBanned(String userProfileId)`, thus the findUserProfile and handling of not found profiles is not needed to replicate here (with the risk of impl a diff. policy).

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -291,19 +327,63 @@ class NodeOffersServiceFacade(
...
-        return true
+
+    private fun isNotIgnored(message: BisqEasyOfferbookMessage): Boolean {
+        val authorUserProfileId = message.authorUserProfileId
+        return !userProfileService.isChatUserIgnored(authorUserProfileId)
+    }
+
+    private fun isTextMessage(message: BisqEasyOfferbookMessage): Boolean {
+        if (message.chatMessageType == ChatMessageType.TEXT) return true
+        return message.text.isPresent && !message.bisqEasyOffer.isPresent
```
HenrikJannsen, line 353: In Bisq 2 main we have 

```
 public static boolean isTextMessage(BisqEasyOfferbookMessage message) {
        return message.getBisqEasyOffer().isEmpty();
    }
```
    
The method name is not good as we do not check for text message but for any non offer message. We have to stick to Bisq 2 implementation to avoid that mobile diverge in behaviour. If we find issues or room for improvements for bisq 2 we should apply it there or open discussion.

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
```diff
@@ -324,9 +404,10 @@ class NodeOffersServiceFacade(
                 channel.market.baseCurrencyName,
                 channel.market.quoteCurrencyName,
             )
+            val count = channel.chatMessages.count { it.hasBisqEasyOffer() && isValidOfferbookMessage(it) }
```
HenrikJannsen, line 407: As we always check for both `it.hasBisqEasyOffer() && isValidOfferbookMessage(it)` we should maybe extract it in a method

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
rodvar: we can do this refactor next time we touch this code

androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/service/offers/NodeOffersServiceFacade.kt
rodvar: I'm not sure what to do with this, feel free to send a follow up PR if you think its worth

rodvar:
rebased to main fixed coflicts and applied suggestions on last pre-merge commit

PR #800, opened 2025-10-04
Add backup and restore feature

### LINKED ISSUE #608: User Profile: Backup/Restore
Image

From the user profile screen, we need to provide the backup & restore identity functionality.

This is important because if a user is invested in a particular profile (buying reputation), without this he has no way to recover his profile if he for example changes device.

For node its about to generate a zip file with the identity files and allow the user to save it (could use OS share functionality so he can choose from all his apps including email)

For connect apps the identity is stored in the trusted node, so when we implement the right profile recovery this is not needed as long as the trusted node admin backups.

HenrikJannsen:
- resolves #608 

This adds a backup and restore functionality. This is important for sellers with reputation to ensure their investment is not lost in case the data loss. Also for developers when testing real trades as sellers, they need to invest in reputation and with that feature that investment is not lost after app uninstall.
Users could also use that to switch between desktop and mobile, though I have not tested that, but backing up mobile and using the zip to replace the data dir of desktop should work without issues.   

### Overview
When doing a backup the user can choose optionally a password. We backup only the 2 relevant directories:
- private (all the keys, private messages, trades)
- settings (user settings, lower prio)

If password is used the file ends with `.enc` otherwise with `.zip`. The user can choose how to share the backup (send, by email, store on drive,...

For restore there is a file browser intent and if the selected file ends with `.enc` we show the password dialogue otherwise the file is processed directly.

Here the steps whats happening:
- If encrypted, user provide password and we decrypt the zip file
- The zip file gets unzipped to the files directory with name `bisq_db_from_backup`
- We shut down all services, delete the `private` and `settings` directories in the `db` directory and trigger a restart
- At startup we check if there is the `bisq_db_from_backup` directory present, if so we move the 2 sub directories to the `db` directory and delete the `bisq_db_from_backup` directory. After that we continue the startup. Now the data from the restored backup is used.

We do not do an extra backup before restore, because we have anyway the automated backups and we tell the user to make a backup before restore just to be safe. The backup data in the filesystem, are anyway not really accessible as far I am aware, thus of little benefit.

This is node only as for client the data are on the backend thus backup does not make much sense.

## Summary by CodeRabbit

- **New Features**
  - In-app backup and restore with optional password encryption, save-to-Downloads, and sharing via system chooser.
  - Guided restore flow with file picker, validation, progress/error overlays, and automatic app restart on successful restore.
  - Settings UI: Backup/Restore controls and password dialog; password field visibility toggle.

- **Localization**
  - Added strings for backup/restore UI, prompts, success, and detailed error messages.

- **Platform**
  - Secure Android file sharing enabled via a FileProvider and exposed cache path for backups.

- **Tests**
  - Unit tests for file utilities, zip/unzip, move/replace behavior, and encryption roundtrip.

- **Chores**
  - iOS project build file cleanups (no user-facing changes).

### DIFF
--- a/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/presentation/NodeResourcesPresenter.kt
+++ b/androidNode/src/androidMain/kotlin/network/bisq/mobile/android/node/presentation/NodeResourcesPresenter.kt
@@ -4,6 +4,9 @@ import android.content.Context
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.delay
 
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
 import network.bisq.mobile.android.node.NodeApplicationLifecycleService
 import network.bisq.mobile.android.node.utils.copyDirectory
 import network.bisq.mobile.domain.utils.decrypt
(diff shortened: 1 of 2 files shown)


rodvar:
rebased to main

PR #844, opened 2025-10-17
chore: small refactor to prep using KmpTorService on client

wodoro:
## Summary by CodeRabbit

* **Refactor**
  * Simplified initialization so app and Tor services start via no-argument calls; dependency wiring now provides storage locations at construction. Improved lifecycle handling for startup, clean shutdown and reset behavior.

* **New Features**
  * Added a public API to retrieve the Tor SOCKS port once available.


rodvar (approved):
ACK

PR #873, opened 2025-10-24
fix: replace select icon with expand all

wodoro:
image

## Summary by CodeRabbit

* **Style**
  * Replaced the dropdown arrow with a new expand indicator across select and multi-select controls for a more consistent, modern look.
  * Added the new expand icon rendering so selection components display the updated indicator throughout the app.


rodvar (approved):
utACK

thanks wod!

PR #884, opened 2025-11-03
bugfix/overlay_not_showing_on_zombie_sstate_reconnect

### LINKED ISSUE #824: Not showing reconnect overlay after long time in background
I saw now 2 times that after very long being in background (> 12 hours) the app has 0 connections, the profile icon has the ? placeholder and after a while it start loading the inventory request and recovers. This is good that it recovers by itself, but it is missing the overlay with the reconnect message and restart button usually shown in that case.

rodvar:
- fix #824 
 - use persistance to avoid edge case on zombie state app not showing overlay when recovering connections/data

## Summary by CodeRabbit

* **New Features**
  * Improved connectivity behavior: reconnection overlay now respects whether the app has synced on the current version, reducing false reconnection prompts after upgrades.
  * Settings now persist an "ever received all data" flag and the last-seen app version to drive upgrade-aware UI logic.

* **Tests**
  * Added unit tests for reconnection overlay logic and updated test fakes to support the new settings flag.

### DIFF
--- a/apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
+++ b/apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
@@ -13,7 +13,8 @@ class NodeConnectivityService(
     private val settingsRepository: SettingsRepository,
 ) : ConnectivityService() {
 
-    private var hasOnceReceivedAllData: Boolean = false
+    // Cached mirror of settings-scoped flag indicating we have fully synced on this app version
+    private var hasSyncedOnThisVersion: Boolean = false
     private var collectJob: Job? = null
 
     // Activated after application service is initialized.
@@ -28,7 +29,7 @@ class NodeConnectivityService(
                 null
             }
             val isFirstRunAfterUpgrade = settings?.lastSeenNodeAppVersion != BuildNodeConfig.APP_VERSION
-            hasOnceReceivedAllData = if (isFirstRunAfterUpgrade) false else (settings?.everReceivedAllData ?: false)
+            hasSyncedOnThisVersion = if (isFirstRunAfterUpgrade) false else (settings?.everReceivedAllData ?: false)
             // Do not update lastSeenNodeAppVersion yet; wait until we have full data in this version
 
             combine(nodeNetworkServiceFacade.numConnections, nodeNetworkServiceFacade.allDataReceived) { numConnections, allDataReceived ->
@@ -36,8 +37,8 @@ class NodeConnectivityService(
             }.collect { (numConnections, allDataReceived) ->
                 // allDataReceived in the network layer will get reset to false when we lose all connections.
                 // We keep whether we've ever seen full data to distinguish reconnect flows
-                if (allDataReceived && !hasOnceReceivedAllData) {
-                    hasOnceReceivedAllData = true
+                if (allDataReceived && !hasSyncedOnThisVersion) {
+                    hasSyncedOnThisVersion = true
                     // Persist flags for this version once we have full data
                     try {
                         settingsRepository.update {
@@ -54,7 +55,7 @@ class NodeConnectivityService(
                 if (numConnections < 0) {
                     _status.value = ConnectivityStatus.BOOTSTRAPPING
                 } else if (numConnections == 0) {
-                    if (hasOnceReceivedAllData) {
+                    if (hasSyncedOnThisVersion) {
                         _status.value = ConnectivityStatus.RECONNECTING
                     } else {
                         _status.value = ConnectivityStatus.DISCONNECTED


apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/android/node/service/network/NodeConnectivityService.kt
```diff
@@ -16,14 +20,32 @@ class NodeConnectivityService(
     override fun activate() {
         collectJob?.cancel()
         collectJob = serviceScope.launch {
+            // Load persisted flag before collecting status changes, with version gating
+            val settings = try {
+                settingsRepository.fetch()
+            } catch (e: Exception) {
+                null
+            }
+            val isFirstRunAfterUpgrade = settings?.lastSeenNodeAppVersion != BuildNodeConfig.APP_VERSION
```
HenrikJannsen: We do you check for the version? The everReceivedAllData is not really relevant to an app version update.

HenrikJannsen (changes requested):
In case hasOnceReceivedAllData is false (as in the case when the OS kills the app), we should get into the ConnectivityStatus.DISCONNECTED state which should show the other popup which is even more appropriate for that case (with restart button).
I suspect the problem is that the ConnectivityStatus is null after app kill. We should add a null check and add a new popup to tell the user that the app **need** to be restarted. 

In fact we need to check that at the first entry point and if the app got killed not allow to proceed.

rodvar:
> In case hasOnceReceivedAllData is false (as in the case when the OS kills the app), we should get into the ConnectivityStatus.DISCONNECTED state which should show the other popup which is even more appropriate for that case (with restart button). I suspect the problem is that the ConnectivityStatus is null after app kill. We should add a null check and add a new popup to tell the user that the app **need** to be restarted.
> 
> In fact we need to check that at the first entry point and if the app got killed not allow to proceed.

ConnectivityStatus is never null (it’s a StateFlow initialised to BOOTSTRAPPING), so the null case can’t explain missing UI.

We already show the correct UI in both OS-kill scenarios:
 - If the app previously fully synced on this version → RECONNECTING → reconnect overlay.
 - If it never synced (fresh start / upgrade / killed mid‑bootstrap) → DISCONNECTED → “all connections lost” dialog with restart button.

Forcing a restart on every OS kill would degrade UX. The current situation as you described in the original issue is that the user brings the zombie to the foreground finding no data is available and his avatar has the default with the question mark.
After a few seconds everything is back up. The goal of this PR is to show the overlay on that transition so the user can clearly see connectivity got lost and we are recovering the app data. 
We should only prompt restart if bootstrap actually fails or the app gets stuck, which we already handle via the bootstrap flow/timeouts (which we already do).

Does this make sense?

rodvar:
> > > In case hasOnceReceivedAllData is false (as in the case when the OS kills the app), we should get into the ConnectivityStatus.DISCONNECTED state which should show the other popup which is even more appropriate for that case (with restart button). I suspect the problem is that the ConnectivityStatus is null after app kill. We should add a null check and add a new popup to tell the user that the app **need** to be restarted.
> > > In fact we need to check that at the first entry point and if the app got killed not allow to proceed.
> > 
> > 
> > ConnectivityStatus is never null (it’s a StateFlow initialised to BOOTSTRAPPING), so the null case can’t explain missing UI.
> > We already show the correct UI in both OS-kill scenarios:
> > ```
> > * If the app previously fully synced on this version → RECONNECTING → reconnect overlay.
> > 
> > * If it never synced (fresh start / upgrade / killed mid‑bootstrap) → DISCONNECTED → “all connections lost” dialog with restart button.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Forcing a restart on every OS kill would degrade UX. The current situation as you described in the original issue is that the user brings the zombie to the foreground finding no data is available and his avatar has the default with the question mark. After a few seconds everything is back up. The goal of this PR is to show the overlay on that transition so the user can clearly see connectivity got lost and we are recovering the app data. We should only prompt restart if bootstrap actually fails or the app gets stuck, which we already handle via the bootstrap flow/timeouts (which we already do).
> > Does this make sense?
> 
> I think that issues just showed a bigger issues which luckily it seems it has not led to user facing bugs, but it certainly will. If the app memory got erased, the app is in an inconsistent state and letting the user continue is asking for troubles. There is lot of important state data in the app and the core library layer. I am surprised that the app could recover at all in such a scenario, but I would not trust on it that there would not be critical bugs from that case. I think we even should not show a popup but automatically trigger a restart. It is not that much different to the loading and reconnecting just that Tor needs to get started up again. We also don't know how well the Tor library deals with kill scenarios.

I think this is a fair concern, we are talking about the most complex situation in Android. There is no lifecycle callback for this.
I've done trading after a recovery from this state before and it works, although I wouldn't dare betting there wouldn't be any other issues..

The fix I'm implementing here should work to show the overlay as requested in #824 (and I actually agree this is desirable to tackle this corner case)

Implementing a proper deterministic restart would imply a proper global zombie mode detection, this can be done by persisting something like a "session token" with the pid of the app and a comparison on application create.

Let's keep this on hold a bit and test a bit more this scenario with the app to see if we can actually find some bugs after the recovery and then decide what is worth?

re your "tor lib", does the overlays for sync make sense for the connect apps since all it needs to do is reconnect to the trusted node..?

PR #891, opened 2025-11-05
fix: dont resume state when app gets killed

### LINKED ISSUE #824: Not showing reconnect overlay after long time in background

wodoro:
- fixes #824 

This will make the app always start from the bootstrap screen no matter how it is killed. if it starts from a dead state, it needs to go through bootstrap instead of jumping straight to the app

you can test the before/after this PR by killing the app and starting it again using this:
```
# Restart ADB daemon as root
adb root

# Now you can use kill commands
adb shell kill -9 $(adb shell pidof network.bisq.mobile.node.debug)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app initialization and recovery when the system terminates the application process, ensuring proper reinitiation on startup.

### DIFF
--- a/apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/android/node/NodeMainActivity.kt
+++ b/apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/android/node/NodeMainActivity.kt
@@ -3,6 +3,7 @@ package network.bisq.mobile.android.node
 import android.os.Bundle
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import network.bisq.mobile.presentation.MainActivity
+import network.bisq.mobile.presentation.MainApplication
 
 /**
  * Bisq Android Node Main Activity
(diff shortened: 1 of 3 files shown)


apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/android/node/NodeMainActivity.kt
```diff
@@ -12,7 +13,13 @@ class NodeMainActivity : MainActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         // Install splash screen before super.onCreate to prevent UI blocking
         installSplashScreen()
-        super.onCreate(savedInstanceState)
+
+        if (MainApplication.wasProcessDead.getAndSet(false)) {
```
rodvar: code dup - could be deletected to MainApplication and call from both activities.

PR #885, opened 2025-11-03
feat: implement auth

wodoro:
this PR implements authentication required by the http api

## Summary by CodeRabbit

* **New Features**
  * Added API password authentication for trusted node connections, a password field in Advanced Options, sensitive settings storage, and platform cryptography/local encryption.

* **Bug Fixes / Reliability**
  * Better handling/reporting for unauthorized/missing-password errors; avoid reconnects for auth failures; enforced 5 MB request body limit and safer request body hashing.

* **User Interface**
  * Simplified onboarding button logic; adjusted Advanced Options layout and proxy control availability.

* **Tests / Chores**
  * Removed obsolete settings setters and updated tests/migrations to match new settings shape.

* **Documentation**
  * Added localization keys for password and auth error messages.

PR #893, opened 2025-11-05
fix: the retry condition

wodoro:
fix for bug introduced in #885 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket reconnection handling to attempt reconnection across more error scenarios, with the exception of explicitly refused connections. This enhances connection stability and reliability across varying network conditions.


rodvar (approved):
utACK

PR #906, opened 2025-11-08
fix: layout.buildDirectory usages

wodoro:
logging the actual string shows incorrect paths, but somehow this was working while it shouldnt
its possible that there's subtle behavior in runtime, or gradle is doing optimistic parsing of paths
either way it seems to me that we are relying on non-deterministic behavior here and that it should be fixed

## Summary by CodeRabbit

* **Chores**
  * Improved build configuration to reliably locate and include generated sources for each build variant.
  * Made variant handling more robust and case-insensitive, with a sensible default for unmapped or unexpected variants to reduce build failures.


wodoro:
it should be fine now

PR #946, opened 2025-11-17
trusted node setup fixes and improvements

### LINKED ISSUE #940: Bootstrap: tor timeout is too shallow for slow connections
Issue created from real world testing against trusted node prod tor auth setup

30 seconds makes the connection attempts to fail before the server is actually able to respond, bringing a bad user experience also when the trusted node was successfully setup, it will failed "connecting trusted node" and show the connection setup screen again.

Let's double this time for prod releases.

rodvar:
- fix #940 
 - bump tor connect timout to 60 seconds
 - main action button becomes cancellation button when testing connectivity.
 - tor: if tor embedded daemon already started, do not kill it.
 - ux: make main action button always visible
 - make system use 8090 (default port) if user doesn't input a port
 - make trusted node setup description more user friendly
 - added unit tests to verify new functionality

## Summary by CodeRabbit

* **New Features**
  * Explicit cancel action for connection attempts; prevents concurrent connects and tracks/cancels countdowns/jobs.

* **Bug Fixes**
  * Tor connection timeout increased from 30s to 60s to reduce Tor endpoint failures.

* **UI**
  * Connect action moved to a bottom bar; button shows "Cancel" during loading and reflects input/validation state.
  * Trusted Node labels and info text updated for clarity.

* **Tests**
  * Added unit tests, test DI module, test utilities and coroutine test tooling.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/startup/TrustedNodeSetupScreen.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/ui/uicases/startup/TrustedNodeSetupScreen.kt
@@ -96,25 +96,53 @@ fun TrustedNodeSetupScreen(isWorkflow: Boolean = true) {
         } else null,
         snackbarHostState = presenter.getSnackState(),
         bottomBar = {
-            Row(
-                horizontalArrangement = Arrangement.Center,
-                modifier = Modifier.fillMaxWidth().padding(vertical = BisqUIConstants.ScreenPadding, horizontal = BisqUIConstants.ScreenPadding)
+            Column(
+                modifier = Modifier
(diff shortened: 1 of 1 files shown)


wodoro:
utACK
sounds good to me

as a side note, I remembered a suggestion made by henrik about tor,
he suggested setting up a keep alive request/response to keep the connection alive
I had forgotten this
may need to look into it, as I also remember seeing something along these lines already implemented, not sure though

rodvar:
@wodoro you are right with both comments, would you like to disregard this PR and do it altogether?

rodvar:
> @wodoro you are right with both comments, would you like to disregard this PR and do it altogether?

I'll do the cancelation button in this PR and merge Wod, and you can take the keep-alive setup for websockets if you will 💪

PR #1030, opened 2025-12-25
fix: remove unnecessary --rerun-tasks from ktlint cmd

wodoro:
according to https://github.com/JLLeitschuh/ktlint-gradle?tab=readme-ov-file#faq , ktlint supports incremental change detection, which should save us some time on the hook execution locally, but ` --rerun-tasks ` forces it to take a long time every time

## Summary by CodeRabbit

* **Chores**
  * Improved build tooling to avoid redundant formatting/check runs, reducing unnecessary task executions and speeding up local verification steps.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -224,7 +224,7 @@ tasks.register<Exec>("ktlintFormatAndCheck") {
     description = "Formats code with ktlint and fails if any violations remain"
     group = "formatting"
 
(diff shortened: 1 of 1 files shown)


rodvar (approved):
ACK

it works well for me as well. Thanks guys, if we found some issue or corner case we can always tweak this in the following days 👍

PR #1051, opened 2026-01-04
Fix robolectric implementation so we can ran ui tests locally

### LINKED ISSUE #796: Enable Robolectric Compose UI test
The UI test in shared/presentation/src/androidUnitTest/.../MessageDeliveryInfoAndroidUiTest.kt is @Ignore due to a Robolectric Compose idling NPE (RobolectricIdlingStrategy.android.kt:32).

 - Goal: run this test under androidUnitTest (no emulator), keep CI fast.

Acceptance:

 - ./gradlew :shared:presentation:test runs the test (not skipped) and passes on macOS and Linux CI.

Notes:

 - Keep: ui-test-junit4 (androidUnitTest), ui-test-manifest (debug).
 - Ensure Robolectric has resources (includeAndroidResources).
 - Consider adding test annotations: AndroidJUnit4 runner, @Config(sdk = 33), @LooperMode(PAUSED).

salottodev:
- fix #796 

## Summary by CodeRabbit

* **Chores**
  * Updated test tooling version and switched to the "prov" variant of the crypto library.

* **Tests**
  * Adjusted test dependencies and scopes to better align unit tests with production.
  * Enabled Android resources and default-return behavior for unit tests.
  * Stabilized UI tests with a centralized setup, switched Android test runner for compatibility, simplified asset handling, and added synchronization before assertions.

### DIFF
--- a/shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/common/ui/components/molecules/chat/MessageDeliveryInfoAndroidUiTest.kt
+++ b/shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/common/ui/components/molecules/chat/MessageDeliveryInfoAndroidUiTest.kt
@@ -48,7 +48,6 @@ class MessageDeliveryInfoAndroidUiTest {
 
     @Test
     fun messageDeliveryInfo_composes_with_non_empty_map_and_shows_text() {
-        I18nSupport.setLanguage("en")
         val info =
             MessageDeliveryInfoVO(
                 messageDeliveryStatus = MessageDeliveryStatusEnum.SENT,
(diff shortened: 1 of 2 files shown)


apps/nodeApp/build.gradle.kts
```diff
@@ -289,6 +289,12 @@ configurations.all {
     exclude(group = "com.github.bisq-network", module = "jsocks")
 }
 
+// Exclude conflicting dependencies from Robolectric to use Bisq core versions
+configurations.matching { it.name.contains("UnitTest") }.configureEach {
+    exclude(group = "org.bouncycastle", module = "bcprov-jdk18on")
+    exclude(group = "com.google.guava", module = "guava")
+}
+
```
rodvar: this PR is all great except this part, this would led to subtle mismatch of bouncycastle and guava versions between prod app and the one used for testing.

Could we consider using dependency resolution commands like force? Something like:

```
// Force Bisq2 core dependency versions in unit tests to match production
// Robolectric 4.16 brings bcprov-jdk18on:1.81, but Bisq2 requires 1.78.1
configurations.matching { it.name.contains("UnitTest") }.configureEach {
    resolutionStrategy {
        force(libs.bouncycastle)
        force(libs.bouncycastle.pg)
        force(libs.google.guava)
    }
}
```

rodvar (changes requested):
only one thing to review/discuss and we can bring this in 👍

rodvar (approved):
ACK

excellent 👍

PR #1069, opened 2026-01-07
bugfix/regression_foreground_not_started_in_time_exception

### LINKED ISSUE #853: android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException
Still happening in some devices.

**Device**: Redmi Redmi Note 11
**Android**:  13 (SDK 33)]
**Version**: 17 (0.1.3)
**Status**: Foreground

```
Exception android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException:
  at android.app.ActivityThread.generateForegroundServiceDidNotStartInTimeException (ActivityThread.java:2124)
  at android.app.ActivityThread.throwRemoteServiceException (ActivityThread.java:2098)
  at android.app.ActivityThread.-$$Nest$mthrowRemoteServiceException (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2360)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loopOnce (Looper.java:211)
  at android.os.Looper.loop (Looper.java:300)
  at android.app.ActivityThread.main (ActivityThread.java:8503)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:561)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:954)
```

rodvar:
- resolves #853 
 - please see explanations in associated issue last comment☝️ 
 - This PR do not allow foreground service init if app is still doing the init sync

## Release Notes

* **Bug Fixes**
  * Improved reliability of the notification service by ensuring it only activates after the application is fully initialized, preventing startup errors.

### DIFF
--- a/apps/clientApp/src/androidMain/kotlin/network/bisq/mobile/client/common/di/ClientDomainModule.android.kt
+++ b/apps/clientApp/src/androidMain/kotlin/network/bisq/mobile/client/common/di/ClientDomainModule.android.kt
@@ -3,6 +3,7 @@ package network.bisq.mobile.client.common.di
 import network.bisq.mobile.client.main.ClientMainActivity
 import network.bisq.mobile.domain.service.AppForegroundController
 import network.bisq.mobile.domain.service.ForegroundDetector
+import network.bisq.mobile.domain.service.bootstrap.ApplicationLifecycleService
 import network.bisq.mobile.presentation.common.notification.ForegroundServiceController
 import network.bisq.mobile.presentation.common.notification.ForegroundServiceControllerImpl
 import network.bisq.mobile.presentation.common.notification.NotificationController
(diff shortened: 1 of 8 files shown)


wodoro:
I like the idea but I'm not sure if that's the best place to put it
my guess is the actual work that causes this finishes once the app has finished receiving network data (when banner turns green). so listening for `NetworkServiceFacade.allDataReceived` may be a better idea ?

again, this is just my guess.

my reasoning is that service activation is not supposed to take long (block) to finish unless it's absolutely necessary, so it should be finishing while there's still work to do

wodoro:
Also we can simply keep the notification at all times (?)

wodoro:
The immediate problem that comes to mind for the solution in this PR is that this makes a somewhat large window on node app in which the app won't launch the foreground service when going into background and if user expect it to keep running and never open the app again after the initial launch, it may become unreliable

rodvar:
> Contributor

Thanks for taking some time to look into this wod ! 

yes this is the strongest reason why I don't want to get it in and I'm thinking of alternatives...
At the moment is not a big deal but if we have several users with low range phones it could be.

keeping the notification on when its not needed is a waste of resources / battery drainer so that's a no go.

Maybe we can leave it as is for now (drop this PR) and try to get some time to help optimise bisq2 if possible to reduce the wait times of sync at least to what it was in prior versions.

Issue #1140, opened 2026-02-09
Minor UI issues


nostrbuddha:
Build version: 41a00788d71cb95e3a5a4e467292303e5fceb38e
Bisq2 version: for-mobile-based-on-2.1.8 (c9be867cb14c4a4e1da2b114fb53bf923e36ee61)

### 1. Sell amount limitation
1. Navigate to Create offer :: Sell :: Amount.
2. User’s reputation allows only a max amount of 250 USD.
3. Fiat Input field and slider works right.
4. But the max label below the slider, reads 600 USD. Should be 250 USD.
5. In desktop, 600 USD is shown, but the slider UI is different, to differentiate between max allowed in app, and allowed as per user reputation.
6. Applicable in both Fixed and Range amount.

<img width="503" height="872" alt="Image" src="https://github.com/user-attachments/assets/c8ad73b3-736f-4ad9-942c-2a56e6659b71" />

### 2. Amount to receive alignment issue.
1. Navigate to Take offer :: Sell :: Review trade
<img width="501" height="850" alt="Image" src="https://github.com/user-attachments/assets/a648c83c-737c-40f5-a42c-044dcf7d13dc" />

### 3. Fixed price is shown as 0
1. Look at an offer with Fixed price in offer book. Price is shown as `@ 0 [CURRENCY]`
1. Take that offer
2. Notice in My trades :: Trade card, that price is shown as 0.
3. Also in Trade screen :: Header section :: Price is shown as 0.

https://github.com/user-attachments/assets/1affedf4-d643-487c-890e-d40735848377

### 4. Create offer :: Fixed price :: % calculation issue
1. Navigate to Create offer :: Fixed price tab
2. Change the fixed price by entering values in the input field.
3. Notice the % value is updating wrongly.

https://github.com/user-attachments/assets/0f62a4fa-d28a-493c-8225-367a972e79ba

### 5. Animations looping in trade screen.
1. On trade state sections, circling animation over the icons, happens only once in desktop.
2. In mobile, it keeps looping.

### 6. After taking offer, clicking on “Show trade in Open trades”, is not navigating to “My trades”.
1. Take a buy offer.
2. On Confirmation dialog, clicking on “Show trade in Open trades”.
3. It navigates back to Offerbook and not switching to 3rd tab, “My trades”.
4. When taking sell offer, this works good.
5. Related: Could rename the button text in Offer take success dialog from "Show trade in Open trades" to “Show trade in My trades”.

[...]

PR #1148, opened 2026-02-10
Fix payment account API calls to use query parameters with proper URI encoding and add connection timeout to prevent indefinite hanging

salottodev:
Depends on: https://github.com/bisq-network/bisq2/pull/4376
Fixes issue 7: https://github.com/bisq-network/bisq-mobile/issues/1140

## Summary by CodeRabbit

* **Bug Fixes**
  * Delete confirmation dialog now closes immediately when confirmed.
  * Improved WebSocket handling with a connection timeout for more reliable requests.
  * Fixed account save/delete request URL formatting and encoding to improve reliability.

* **Validation Updates**
  * Payment account name: min 2 characters, max 20 characters.
  * Payment account description: min 3 characters, max 1000 characters.


rodvar (changes requested):
Overall it works really well, well done. The comment on the PR coverage was just a comment to explain the failure in the PR logs, its ok to leave it at 40% for this case because most of those lines are moved lines from existed untested code.

Just some minor requests to merge this one:

 - the mobile.user.paymentAccounts.createAccount.validations.name.maxLength seems to not be used?
 - See screenshots below, its giving always a generic error for each case so users might have no idea why its failing

Image
Image
Image
Image

PR #1169, opened 2026-02-17
Fix Transifex locale upload no-op for country locales

hiciefte:
## Summary
Fixes a silent Transifex upload no-op for country-specific locales in the mobile sync workflow.

## Root Cause
The batching script converted locale codes like pt_BR/af_ZA to hyphen format (pt-BR/af-ZA). In this repository setup, tx push -t -l pt-BR exits with code 0 but uploads nothing.

## Changes
- Keep matrix locales in repository/.tx format (no underscore->hyphen conversion).
- Use matrix locale keys as-is when mapping locale->resources in JSON output.
- Add workflow guard: fail if tx push exits 0 but no [locale] - Uploading file entries are present.
- Add regression tests for locale preservation and resource mapping.

## Validation
- python3 -m unittest scripts/test_generate_transifex_batches.py
- Local forced push probes confirmed:
  - pt-BR and af-ZA: 0 uploads (no-op)
  - pt_BR and af_ZA: expected uploads

## Summary by CodeRabbit

* **Tests**
  * Added unit tests to validate translation batch generation and resource emission across mixed locale formats.

* **Chores**
  * Improved validation in the translation sync workflow to detect and report incomplete or zero-file uploads per locale.
  * Preserved repository locale formatting in batch outputs to ensure consistent locale handling.

### DIFF
--- a/scripts/generate_transifex_batches.py
+++ b/scripts/generate_transifex_batches.py
@@ -36,15 +36,6 @@ PRIORITY_TIERS["standard"] = [
 ]
 
 
-def convert_locale_format(locale: str) -> str:
-    """
-    Keep locale codes in repository/.tx format for tx CLI.
-
-    Examples: pt_BR, af_ZA.
-    """
-    return locale
-
-
 def generate_batches(batch_size: int, priority_based: bool = False, locales: Optional[List[str]] = None) -> List[Dict]:
     """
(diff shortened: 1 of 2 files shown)


rodvar (approved):
utACK

merging to test CI

Issue #1171, opened 2026-02-18
Persist sortby across sessions


rodvar:
### 8. Offerbook sortby suggestion:

Would be good to have Offerbook Sort by, Show markets selection persist between app runs.

as extracted from #1140

PR #1179, opened 2026-02-23
feat: persist market SortBy and market filter & add tests

wodoro:
resolves https://github.com/bisq-network/bisq-mobile/issues/1171

I added tests using AI, and refactored duplicate code into shared files
refactored ComputeOfferbookMarketListUseCase out of OfferbookMarketPresenter and added tests for it

## Release Notes

* **New Features**
  * Added market sorting capabilities with three options: Most Offers, Name (A-Z), and Name (Z-A)
  * Added market filtering with two options: Show Only Markets with Offers or Show All Markets
  * Market preferences are now automatically saved and persist across sessions

### DIFF
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/datastore/DataStore.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/data/datastore/DataStore.kt
@@ -36,4 +36,5 @@ val dataStoreJson =
         ignoreUnknownKeys = true
         encodeDefaults = true
         explicitNulls = false
+        coerceInputValues = true
     }


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/offers/usecase/ComputeOfferbookMarketListUseCase.kt
```diff
@@ -0,0 +1,59 @@
...
+import network.bisq.mobile.domain.data.model.offerbook.MarketListItem
+import network.bisq.mobile.domain.service.market_price.MarketPriceServiceFacade
+import network.bisq.mobile.domain.utils.CurrencyUtils
+import network.bisq.mobile.domain.utils.Logging
+import network.bisq.mobile.presentation.tabs.offers.MarketFilterUtil
+
+class ComputeOfferbookMarketListUseCase(
+    private val marketPriceServiceFacade: MarketPriceServiceFacade,
+) : Logging {
+    operator fun invoke(
```
salottodev: should we do this way invoke() or execute()? please let's agreed on the way of doing things, because then we all end up doing in our own and the code is not consistent.

salottodev: Let's do this way, please refactor the trusted node one.

rodvar:
rebased to main

rodvar (approved):
ACK

migration issue described above is fixed - well done wod 💪

PR #1252, opened 2026-03-18
Refactor domain module structure

salottodev:
To check this PR, I would pull the branch and check the structure of the domain module.


rodvar (changes requested):
The intent (clean domain/data separation) is good, but the PR implementation itself is confusing and a huge blast to just accept... I recommend to either:                             
  
  - A) Actually create a separate :shared:data Gradle module with compile-time dependency enforcement (domain depends on nothing, data depends on domain). This justifies the 500+-file churn. OR                                                   
  - B) If the goal is just package reorganization its ok 👍 , do it incrementally in smaller PRs (e.g., first move replicated VOs/Dtos, then services, then repositories) to make it easier to analyse/merge?               
  
Either way, the PR needs a proper description explaining the architectural rationale, and the VO/Dto naming inconsistency should be resolved because it looks like is half way as is? 

 Some types were renamed from VO→Dto (FiatAccountVO→FiatAccountDto) but most VOs kept their names. The criteria for which became Dto and which stayed VO is unclear, making the naming convention harder to follow rather than easier
Please help with this one

PR #1275, opened 2026-03-23
Add payment accounts dto and prepare scaffold to implement all accounts

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added broad support for many new fiat payment rails and account types (ACH, SEPA/SEPA Instant, SWIFT, PIX, Zelle, Revolut, Wise, Faster Payments, and additional regional rails), enabling more ways to add and store payment details.

* **Refactor**
  * Account handling migrated to domain models with improved mapping, serialization and currency-validation — more reliable account sync, deserialization, and UI integration.


rodvar (approved):
utACK

a couple of considerations/questions for the next PR:

1. is CASH_APP missing from serializer (potential crash)? FiatPaymentRailDto has CASH_APP but there's no CashAppAccountDto, no serializer case, and no domain model. If the backend sends a CASH_APP account, FiatAccountDtoSerializer hits the else branch and throws IllegalArgumentException. (is this intentional?)
2. FiatAccountMapping.toDomain() only handles UserDefinedFiatAccountDto - All other account types fall through to else -> throw. The service facade calls toDomain() on every account from the API —  if the backend returns any non-CUSTOM account, the client crashes. This is probably intentional ("prepare scaffold") just double checking

PR #1290, opened 2026-03-26
bugfix/topic_subscription_failure_causes_rest_of_subscriptions_to_be_aborted

rodvar:
- handle per-topic subscription error and proper logging to avoid one failure to affect the rest of the subscriptions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for subscription operations. Individual subscription failures no longer prevent other subscriptions from being applied successfully, enhancing overall system resilience.


wodoro:
imo we need to do all-or-nothing style of subscriptions and either retry or propagate error to the user
letting one sub to succeed and others to fail is a setup for surprises

rodvar:
> imo we need to do all-or-nothing style of subscriptions and either retry or propagate error to the user letting one sub to succeed and others to fail is a setup for surprises

@wodoro  very good point wod, I got hurried into fixing the situation where a topic is not in the backend without thinking of it - thanks

I think we could merge this adding the CRAI proposal in and then work on a UI facing solution to inform the user if there subscriptions failing in a user friendly language?

PR #1299, opened 2026-03-29
"Don't show again" in Open weblink Dialog

### LINKED ISSUE #731: Open weblink should have "Don't show again"
Image

nostrbuddha:
1. Fixes https://github.com/bisq-network/bisq-mobile/issues/731
2. Don't show again selection persisted in SettingsService.dontShowAgain map. Key: DontShowAgainKey.HYPERLINKS_OPEN_IN_BROWSER
3. Should open URL (or) Copy link choice persisted in SettingsService.cookie. Key: CookieKey.PERMIT_OPENING_BROWSER
4. Done for both androidNode and xClients.
5. xClients require API update via bisq2 PR: https://github.com/bisq-network/bisq2/pull/4587

## Summary by CodeRabbit

* **New Features**
  * Web-link confirmation now presenter-driven with “don’t show again” and explicit confirm/dismiss flows.
  * Option to permit auto-opening in browser or copy-to-clipboard, persisted per user choice.
  * Confirmation dialog: per-button loading states, extra content slot, and close button respected when disabled/loading.
  * Settings: “Reset don't show again” button to restore flags with feedback.

* **Tests**
  * Expanded unit and UI tests covering dialog flows, persistence, error handling and multi-step scenarios.

* **Localization**
  * Added success message for resetting flags.


rodvar (changes requested):
Hey buddha welcome back and thanks for sending a contribution after a long while.

This one has several things that we would need to address before merging.

**The positives**:

 - Well done complying with coverage 💪 
 - Most architectural decision were followed (besides the concerns below)
 - Data layer conventions ✅ 

**What needs more work:**

 1. Architecture: the "cookie" concept is not worth it for a preference that should be device linked. Making the user wait for a network request for this preference (and all the concerns one needs to have for that) doesn't make sense either, I would suggest we drop the bisq2 PR and make the necessary changes to use local storage (please check other local settings in repository)
 2. Architecure (UI):  WebLinkConfirmationDialog.kt now directly injecs SettingsServiceFacade and MainPresenter via koinInject() and contains persistence logic (persistWebLinkDialogChoice, copyLinkWithUserFeedback). Per the project's MVP pattern, this logic should live in a presenter, not in a Composable. Views should be stateless; the LaunchedEffect with conditional navigation/clipboard/snackbar is essentially presenter-level orchestration. 
 3. please review Nullable force-access on settingsServiceFacade, could certainly be crash-prone
 4. @ExcludeFromCoverage on entire SettingsApiGateway - please add the exclude only to the methods targetted (probably reified ?)
 5. SettingsPresenter.kt line ~2062: // TODO:i18n Reset flags successfully — uses a non-existent i18n key "mobile.settings.resetFlagsSuccess"
 6. The new methods in SettingsServiceFacade have default implementations returning Result.success(Unit):                         
 ```
  suspend fun setWebLinkDontShowAgain(): Result<Unit> = Result.success(Unit)                                                  
  suspend fun resetAllDontShowAgainFlags(): Result<Unit> = Result.success(Unit)                                                 
  suspend fun setPermitOpeningBrowser(value: Boolean): Result<Unit> = Result.success(Unit)
```                                
  But showWebLinkConfirmation and permitOpeningBrowser are val with no defaults, which forces all implementors to provide them. This inconsistency means the interface can't be partially implemented — the defaults on the functions give a false sense of optionality.  
 7. as already chatted, CRAI comments (including unnecessary change in node devMode)

hope this helps, please let me know if you have any questions

**Review by rodvar — CHANGES_REQUESTED** [review]
good second pass, most structural issues have been addressed.

Please review my code comments (most important) and also:

 - Dual storage (backend + local) but local appears unused — dead code? please cleanup leftovers of the previous approach 🙏

**Review by rodvar — APPROVED** [review]
ACK (all 3 apps)

All major review changes have been address 💪 
Some minor nits (can be sent in a follow up)

  - Dead code: unregisterFromParent() added to BasePresenter but never called - should be removed or justified
  - Missing exception handling: openUri()/copyToClipboard() coroutines don't guard hideLoading() against unexpected exceptions — could leave loading overlay stuck, risk is minor but possible                                                                                                  
  - Brief loading flash: auto-handle path (no dialog) calls showLoading() unnecessarily ?
  - I think the dialog now has a little bit too much padding around the new "don't show again" option? please review

File: shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/settings/SettingsScreen.kt [hunk 1]
```diff
@@ -199,6 +201,14 @@ fun SettingsContent(
                         onSwitch = { onAction(SettingsUiAction.OnUseAnimationsChange(it)) },
                     )
 
+                    BisqGap.V1()
+                    BisqButton(
+                        text = "settings.display.resetDontShowAgain".i18n(),
+                        onClick = { onAction(SettingsUiAction.OnResetAllDontShowAgainClick) },
+                        type = BisqButtonType.GreyOutline,
```
salottodev, line 208: @nostrbuddha @rodvar is this button grey outlined on purpose? it seems like a disabled button.

Image

nostrbuddha: Yes. In bisq2 desktop, it's like this grey outline button. Just following the same style.
I think a disabled button will have it's background color in a different shade of gray.

rodvar: @nostrbuddha correct, but @salottodev is right to point this out - the mobile apps only use grey for disabled comms it could be confusing for our users.

Can you please have a look at the 2 buttons in the Settings -> Backup & Restore section? - I believe that should be used if agreed please send the follow up.

PR #1303, opened 2026-03-30
enhancement/use_factory_for_already_stateless_presenters

rodvar:
- switch instantiation method single->factory for presenters that are already stateless
 - make some low hanging fruit presenters stateless and switch their instantiation method too (UserProfile & CreateProfile)

## Summary by CodeRabbit

* **Refactor**
  * Presenters now instantiate per use instead of reusing shared instances, reducing cross-screen state leakage.
* **Behavior**
  * Background tasks are no longer force-cancelled on view detach; long-running operations may continue after navigation.
  * Entered payment proof/data may persist when a view is detached instead of being cleared automatically.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/di/PresentationModule.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/di/PresentationModule.kt
@@ -113,7 +113,7 @@ val presentationModule =
             )
         }
 
-        single {
+        factory {
             CreateProfilePresenter(
                 get(),
                 get(),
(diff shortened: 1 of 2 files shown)


rodvar:
Remaining singletons (intentionally kept in this PR):
                                                                                    
  1. MainPresenter / ClientMainPresenter / NodeMainPresenter — root of presenter hierarchy -> These are the only presenters that should be singletons by design.                                
  2. OfferbookMarketPresenter / ClientOfferbookPresenter / Node OfferbookPresenter — intentionally preserves filter/sort state across tab switches (this should be refactored before converting to factory instance)                                                                                                
  3. CreateOfferPresenter / TakeOfferPresenter — shared data containers for multi-step wizards -> these needs a refactor its not even a presenter, its used as a repository/coordinator for the workflows.            

@coderabbitai could you please create issues for items 2 and 3 to address in the near future?

PR #1311, opened 2026-03-31
fix: Sell offer: Min range amount in not saved when navigating back and froth

### LINKED ISSUE #1201: Sell offer: Min range amount in not saved when navigating back and froth
Create offer flow (Sell offer) :: Min range amount in not saved when navigating back and froth. It resets to $6.
It works good for buy offer. Happens only with Sell offer.

Video:
https://limewire.com/d/inavO#Rseghnxjse

as reported in #1200

nostrbuddha:
- fixes https://github.com/bisq-network/bisq-mobile/issues/1201

Issue:
 - updateSellerAmountLimitInfo() resets the min slider min range value, when called from init. Since firstLoad is always set to true (updateAmountLimitInfo(firstLoad = true)).
 - This resets the value selected by user, when navigating between screens.

Fix:
 - Conditionally set firstLoad to true or false, based on if the amount value is already set or not.
 - If amount value is already set, then firstLoad is set to false. Otherwise, it's set to true

Test
 - Created a test case, where the CreateOfferAmountPresenter is created twice (to simulate nav back and forth) and the min range is tested to be retaining it value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed offer creation to preserve saved range amount limits when the screen is recreated, ensuring seller-configured minimum and maximum amounts are retained instead of resetting to default values.

* **Tests**
  * Added test coverage for range amount persistence during presenter recreation.

### DIFF
--- a/shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/offer/create_offer/CreateOfferAmountPresenterTest.kt
+++ b/shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/offer/create_offer/CreateOfferAmountPresenterTest.kt
@@ -1,10 +1,12 @@
 package network.bisq.mobile.presentation.offer.create_offer
 
+import io.mockk.coEvery
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.mockkStatic
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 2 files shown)


rodvar (approved):
ACK

neat well tested fix - thanks buddha and welcome back

PR #1325, opened 2026-04-03
feat: implement user-facing sub failure dialog

### LINKED ISSUE #1291: Better error handling for WebSocket topic subscription failures with user-facing feedback and retry
## Background

While investigating a bug where a failure in one topic subscription caused the rest of the subscriptions to be aborted (see PR #1290), it was identified that simply catching and logging individual subscription errors is not sufficient.

A broader, more robust solution is needed.

## Problem

When one or more WebSocket topic subscriptions fail (e.g. because a topic is not available in the backend), the user is not informed about the failure or its impact. This can silently leave the app in a degraded state where certain functionalities are missing with no explanation.

The current approach (silently skipping failed subscriptions) was flagged as a setup for surprises, since some subscriptions succeed and others fail without any user awareness.

## Proposed Solution

Implement a user-facing error handling mechanism for subscription failures that:

1. **Informs the user** which specific functionalities will be unavailable due to failed subscriptions.
2. **Displays a dialog or notification** listing the affected features in user-friendly language (not raw topic names).
3. **Provides a retry action** so the user can attempt to re-subscribe to the failing topics (exact UX flow to be analysed).
4. Consider an **all-or-nothing** strategy vs. partial failure handling – evaluate whether it makes more sense to retry all subscriptions or only the failed ones.

## Related

- PR #1290: https://github.com/bisq-network/bisq-mobile/pull/1290
- Raised by @wodoro and @rodvar in the discussion on PR #1290

## Notes

- The retry mechanism and exact UX flow are to be analysed before implementation.
- This ticket should be scoped to include both the subscription management logic and the UI layer.

wodoro:
resolves #1291

## Summary by CodeRabbit

* **New Features**
  * Non-dismissible dialog showing failed subscription topics (grouped by severity) with Retry and Continue; users may continue with limited features.
  * New connectivity status "Connected with Limitations" that influences startup navigation and flow.

* **UI / Localization**
  * Added localized titles/descriptions for topics and dialog copy.
  * Status indicator content descriptions updated.

* **Tests**
  * Added/updated unit tests covering subscription readiness, failures, and related navigation.

### DIFF
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/trusted_node_setup/TrustedNodeSetupPresenter.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/trusted_node_setup/TrustedNodeSetupPresenter.kt
@@ -103,7 +103,12 @@ class TrustedNodeSetupPresenter(
                     }
 
                     else -> {
-                        _uiState.update { it.copy(status = state.connectionStatus) }
+                        _uiState.update {
+                            it.copy(
+                                status = state.connectionStatus,
+                                serverVersion = state.serverVersion,
+                            )
+                        }
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


wodoro:
I've implemented the hint, but it would require the pairing API to return server version on success too, so a followup pr on bisq2 side then bisq-mobile is necessary

PR #1343, opened 2026-04-09
test: add test for OpenTradeScreen presenters

wodoro:
a quick regression test for #1341

## Summary by CodeRabbit

* **Tests**
  * Added regression tests for trade detail screen presenter instance validation.


wodoro:
it needs to actually test it properly, reviewing the code

wodoro:
I think this is incorrect and this is not doable until we actually can setup a test to run the checks on UI itself and expect the results

rodvar:
it makes sense, thanks for checking though wod

PR #1345, opened 2026-04-10
chore: migrate to the new Preview annotation

wodoro:
since the old Preview is now deprecated and it's [recommended](https://kotlinlang.org/docs/multiplatform/compose-previews.html#available-annotations) to use the new one, I made this PR

## Summary by CodeRabbit

* **Chores**
  * Updated Compose UI preview dependencies to improve development tooling support across the codebase.
  * Removed unused imports from several files.


rodvar:
thanks so much for looking into this wod
When I did the migration I left that for later cause I found some issues reported. 
Will test your PR during this week 🙏

wodoro:
My IDE did not produce error in cases I looked into (I did not check everything), and it was producing previews like before, and the app compiles
if you can tell me what specific issues you had I will look into it before you test this

rodvar (approved):
ACK

Tested on `Android Studio Narwhal 3 Feature Drop | 2025.1.3`

Couldn't find any issues, on the contrary it loads faster than the previous @Preview :)

PR #1361, opened 2026-04-15
fix: resolve build warnings

wodoro:
resolves the remaining build warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed chargeback risk badge null-safety so it no longer crashes when account data is missing.

* **Refactor**
  * Made presenter initialization more robust to reduce runtime issues.
  * Improved offer market processing for clearer internal data handling.

* **Style**
  * Updated reputation screen link styling and interaction for clearer, consistent clickable links.

### DIFF
--- a/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/offers/OfferbookMarketPresenter.kt
+++ b/shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/offers/OfferbookMarketPresenter.kt
@@ -4,7 +4,6 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
-import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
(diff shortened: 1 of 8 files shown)


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/offers/OfferbookMarketPresenter.kt
```diff
@@ -105,22 +104,17 @@ class OfferbookMarketPresenter(
...
-            ) { values: Array<Any?> ->
-                @Suppress("UNCHECKED_CAST")
-                val filter = values[0] as MarketFilter
-                val searchText = values[1] as String
-                val sortBy = values[2] as MarketSortBy
-                val languageCode = values[4] as String
-                val items = values[5] as List<MarketListItem>
-                computeOfferbookMarketListUseCase(filter, searchText, sortBy, languageCode, items)
-            }.flowOn(Dispatchers.Default)
+            network.bisq.mobile.domain.utils
```
rodvar, line 107: nit - we prefer imports to absolute naming

rodvar (approved):
ACK

For node app - build clean of warnings

For clients:

Also resolves many, some pendings I found,  I try to categorize them here:

 **Source warnings not addressed (could be in this PR or follow-up)**:                                                             
  - AlertTypeDtoMapping.kt:13 — 'else' is redundant (same pattern as TradeStatesProvider fix — enum when is exhaustive)
  - ClientUserProfileServiceFacade.kt:169 — decodeBase64Bytes() deprecated, use Base64.Default.decode()                         
  - BackHandler.ios.kt:5, 14 — LocalUIViewController moved to androidx.compose.ui.uikit. Just need to update the import.
                                                                                                                                
 **Out of project scope (third-party / build tooling)**:                                                                           
  - shared/kscan/* — third-party QR scanner module                  
  - Unable to strip libtor.so, libbarhopper_v3.so, ... — NDK build tooling warnings for prebuilt native libs. Harmless I think.      
  - KLIB resolver: same 'unique_name=...' found in more than one library — dependency version mismatch between androidx.compose.* and org.jetbrains.compose.*. Would need to align versions in gradle/libs.versions.toml. Not trivial I believe, I'll create an issue for this one.

rodvar:
- KLIB resolver: same 'unique_name=...' found in more than one library — dependency version mismatch between androidx.compose.* and org.jetbrains.compose.*. Would need to align versions in gradle/libs.versions.toml. Not trivial I believe, I'll create an issue for this one.
  
@coderabbitai can you please create an issue to address this in the future?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/offers/OfferbookMarketPresenter.kt
wodoro: so do I
It seems I forgot to fix after I tested it :(
will do a follow up with the remaining issues

PR #1373, opened 2026-04-20
Update fiat account implementation to user defined

salottodev:
## Summary by CodeRabbit

* **Refactor**
  * Migrated payment account service layer from fiat-specific to user-defined account models. Updated account selection default behavior and internal service facades to support broader account type handling.


rodvar (approved):
utACK

some minor comments to consider for next PR:

1) PaymentAccountsPresenter.observeAccounts() still does state.accounts.filterIsInstance<UserDefinedFiatAccount>(). But AccountsState.accounts is now typed as List<UserDefinedFiatAccount> — the filter is a no-op. Same in SellerState1Presenter - These can be cleaned up?

2) New files AddUserDefinedAccountRequest, SaveUserDefinedAccountRequest, SetSelectedUserDefinedAccountRequest are created but they're structurally identical to the deleted fiat versions — just wrapping a UserDefinedFiatAccountDto. The PaymentAccountsApiGateway.saveAccount() was simplified to not use a wrapper, but the user-defined gateway still wraps. Consider aligning the approach (either all wrap or none wrap)? 
  
3) Just checking UserDefinedPaymentAccountsApiGateway uses same base path as PaymentAccountsApiGateway - is that on purpose?

rodvar:
ignoring PR check for this mechanical refactor

PR #1389, opened 2026-04-28
feat: Implement trade history

### LINKED ISSUE #1266: Trade History
Image
Image
Image

Please see design docs for description on how it should be implemented based on the designs.

### LINKED ISSUE #1398: Audit and remove sensitive identity/key material from BisqEasyTradeDto
## Summary

`BisqEasyTradeDto` currently carries sensitive identity/key material (private keys, identity fields) over the wire in the trade list API response. These fields are not consumed by known clients but represent an unnecessary security exposure.

## Problem

- `TradeItemPresentationDto` (used by the open-trades list endpoint) embeds the full `BisqEasyTradeDto`, which includes contract-level and identity-level sensitive data.
- Sending private key material over the wire—even to trusted nodes—is a security risk that should be addressed.
- Fields identified as heavy/sensitive should be audited and either removed or made nullable/optional in the wire shape.

## Proposed Fix

1. Audit all fields in `BisqEasyTradeDto` for sensitivity (private keys, identity material, etc.).
2. Remove or stop populating identified sensitive fields (additive-compatible: make them nullable first, stop populating from bisq2 side on a coordinated release).
3. Optionally mark heavy fields `@Transient` on the mobile side to drop memory overhead immediately.

> **Note:** This is a breaking change if fields are hard-removed. Coordinate with bisq2 desktop release timeline before hard removal. As an interim step, fields can be made nullable in the wire shape and depopulated server-side once a new desktop version is released.

## References

- bisq-network/bisq2#4700 (PR where this was identified)
- Discussion: https://github.com/bisq-network/bisq2/pull/4700#discussion_r3158039186
- Requested by: @wodoro

wodoro:
- resolves #1266 
 - resolves #1398 
 - requires [bisq2 branch](https://github.com/bisq-network/bisq2/pull/4700) to be merged

## Summary by CodeRabbit

* **New Features**
  * Paginated, searchable Trade History with sorting, outcome/role filters, detailed closed-trade view, and closed-trades change signal.
  * My Trades reorganized into Open Trades + My Trades (history) tabs with paging, controlled filter sheets, and trade details dialog.
  * Backend capability probing to gate history availability; notifications deep-link to My Trades (Open).

* **Bug Fixes / Improvements**
  * Safer CSV escaping, consistent fiat scaling/formatting, medium date format, UI refinements (star ratings, payment displays).

* **Tests**
  * Added unit and paging tests for filtering, pagination, mappings, outcome mapping, and capability detection.


shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/replicated/presentation/open_trades/TradeItemPresentationDto.kt
```diff
@@ -29,10 +29,8 @@ data class TradeItemPresentationDto(
     val formattedQuoteAmount: String,
     val bitcoinSettlementMethod: String,
     val bitcoinSettlementMethodDisplayString: String,
-    val bitcoinSettlementMethodCsvDisplayString: String? = null, // bisq2 PaymentMethodSpec.displayString
```
wodoro, line 32: Server never sent these, so why there's difference between node app and client app for this and fiatPaymentMethodCsvDisplayString?

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/mapping/TradeItemPresentationDtoFactory.kt
```diff
@@ -72,35 +70,33 @@ object TradeItemPresentationDtoFactory {
             takerUserProfile = myUserProfileVO
         }
 
-        val mediatorUserProfile: UserProfileVO? = if (contractVO.mediator != null) contractVO.mediator!! else null
+        val mediatorUserProfile: UserProfileVO? = contractVO.mediator
         val peersReputationScore = reputationService.getReputationScore(peersUserProfile.id)
         val peersRReputationScoreVO = Mappings.ReputationScoreMapping.fromBisq2Model(peersReputationScore)
 
         return TradeItemPresentationDto(
```
wodoro, line 77: small cleanup refactor, nothing is changed basically

wodoro:
see https://github.com/bisq-network/bisq2/pull/4700#discussion_r3185333534

rodvar:
rebased to main

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
```diff
@@ -198,6 +248,14 @@ class ClientTradesServiceFacade(
         }
     }
 
+    @Suppress("UNUSED_PARAMETER")
+    private fun handleClosedTradesChange(
+        payload: List<ClosedTradeListItemDto>,
+        modificationType: ModificationType,
+    ) {
+        _closedTradesChangeTick.update { it + 1 }
```
rodvar, line 276: Closed trade addition triggers a full re-fetch instead of an in-place insert. Acceptable for closed trades

Maybe improvement would be to extract the WS payload (instead of discard it), merge into the paging buffer at the right position. I believe standard `androidx.paging` recipe is **RemoteMediator**, but that's a bigger change... For now, document the design choice and later on we can revisit?

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
```diff
@@ -393,6 +412,83 @@ class NodeTradesServiceFacade(
...
+
+                filtered =
+                    when (sortBy) {
+                        TradeSort.OLDEST_FIRST -> filtered.sortedBy { it.tradeCompletedDate ?: it.takeOfferDate }
+                        TradeSort.AMOUNT_HIGH_LOW -> filtered.sortedByDescending { it.quoteAmount }
+                        TradeSort.AMOUNT_LOW_HIGH -> filtered.sortedBy { it.quoteAmount }
+                        TradeSort.NEWEST_FIRST, null -> filtered.sortedByDescending { it.tradeCompletedDate ?: it.takeOfferDate }
+                    }
+
+                // Node runs in-process with bisq2 lib; closed trades already in memory.
```
rodvar, line 453: 👍

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
```diff
@@ -393,6 +412,83 @@ class NodeTradesServiceFacade(
...
+                // Mapping/filter/sort is O(n) regardless, so return everything as a single page.
+                val total = filtered.size
+                Result.success(
+                    PaginatedResponse(
+                        items = filtered,
+                        page = 1,
+                        pageSize = total,
+                        totalItems = total.toLong(),
+                        totalPages = 1,
+                    ),
```
rodvar, line 463: despite the correct comment above, could we please consider respect params.page and params.pageSize even on Node — same logic, just subList at the end?

A power-user could have thousands of closed trade at a point in time and that small improvement could be meaningful.

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
```diff
@@ -585,6 +681,17 @@ class NodeTradesServiceFacade(
...
         val tradeItemPresentationVO = TradeItemPresentationDtoFactory.create(trade, channel, userProfileService, reputationService)
         val openTradeItem = TradeItemPresentationModel.from(tradeItemPresentationVO)
+
+        // If already in a final state (e.g. app restarted with trade closed), add directly to closed.
+        val currentState = Mappings.BisqEasyTradeStateMapping.fromBisq2Model(trade.tradeState)
+        if (currentState.isFinalState) {
+            // Trade is already closed (e.g. app restart with a closed trade). No need to add it to
+            // openTradeItems. Bump the tick so consumers refresh closed-trade views; the actual
+            // list is read from bisqEasyTradeService.closedTrades via getClosedTradesPaginated.
+            bumpClosedTradesTick()
```
rodvar, line 693: not sure I underestand this one..  Does it also bump on a real trade transition into a closed state?

Haven't tested yet but from code are we risking -> 
 
- User closes a trade in-session → tick doesn't fire → paging doesn't invalidate → user has to navigate away and back to see it appear

?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/usecase/trade/FilterOpenTradesUseCase.kt
```diff
@@ -0,0 +1,51 @@
+package network.bisq.mobile.domain.usecase.trade
+
+import network.bisq.mobile.data.replicated.presentation.open_trades.TradeItemPresentationModel
+import network.bisq.mobile.domain.model.trade.TradeRoleFilter
+import network.bisq.mobile.domain.model.trade.TradeSort
+
+class FilterOpenTradesUseCase {
```
rodvar, line 7: 👏 

will test this throughfully on my next pass

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
```diff
@@ -393,6 +412,83 @@ class NodeTradesServiceFacade(
         _selectedTrade.value = null
     }
 
+    override suspend fun getClosedTradesPaginated(
```
rodvar, line 415: could we please add tests like **ClosedTradesPagingSourceTest** and a **NodeTradesServiceFacadeClosedTradesTest**

can be assited by AI 👍

rodvar (changes requested):
first code-only review pass

overall looking very solid wod, please see my comments in the code.

Once we finish this first review I'll do a full on end2end test

rodvar:
CI failing due to flaky test (unrelated to this PR)

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/replicated/presentation/open_trades/TradeItemPresentationDto.kt
rodvar: > Server never sent these, so why there's difference between node app and client app for this and fiatPaymentMethodCsvDisplayString?

this exists for a stale bisq2 historical reason, I think its safe to remove wod.

connect apps always see null in those files and fallback to the short ones because it has never been wired up to the api.

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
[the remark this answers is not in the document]
wodoro: the problem is node data is not paginated, so paginating it will cause filter and sort on every page load which is unnecessary. it is already commented.

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/replicated/presentation/open_trades/TradeItemPresentationModel.kt
```diff
@@ -34,25 +38,21 @@ data class TradeItemPresentationModel(
...
-        get() {
-            val btc =
-                tradeItemPresentationDto.bitcoinSettlementMethodCsvDisplayString
-                    .takeUnless { it.isNullOrEmpty() } ?: bitcoinSettlementMethodDisplayString
-            val fiat =
-                tradeItemPresentationDto.fiatPaymentMethodCsvDisplayString
-                    .takeUnless { it.isNullOrEmpty() } ?: fiatPaymentMethodDisplayString
-            return "$btc / $fiat"
-        }
+        get() = "$bitcoinSettlementMethodDisplayString / $fiatPaymentMethodDisplayString"
```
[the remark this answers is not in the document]
wodoro: csv variants are removed, so it doesnt matter. the same problem existed before, we basically changed nothing ?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/my_trades/MyTradesScreen.kt
```diff
@@ -0,0 +1,59 @@
...
+    val uiState by presenter.uiState.collectAsState()
+    val isInteractive by presenter.isInteractive.collectAsState()
+
+    val tabOptions =
+        remember {
+            listOf(
+                "mobile.myTrades.tab.open".i18n(),
+                "mobile.myTrades.tab.history".i18n(),
+            )
+        }
```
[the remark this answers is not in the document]
wodoro: so, why not use a language key for remember?

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
wodoro: the reasoning is that we don't know where in the sorting it is placed. simply reloading avoids buggy/incorrect positioning
sounds like RemoteMediator usecase is for when we implement a local db
will add comment

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
wodoro: I did not understand, can you clarify please?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/my_trades/MyTradesScreen.kt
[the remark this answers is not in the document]
wodoro: Maybe it's a good idea to create an issue to discuss providing the language code as a local provider up in the tree? for now I'll remove the remember

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/mapping/trade/BisqEasyClosedTradeMapping.kt
```diff
@@ -0,0 +1,37 @@
...
+        isMaker = trade.isMaker,
+        isBuyer = trade.isBuyer,
+        outcome = Mappings.BisqEasyTradeStateMapping.fromBisq2Model(trade.tradeState).toTradeOutcome(),
+        takeOfferDate = contract.takeOfferDate,
+        tradeCompletedDate = trade.tradeCompletedDate.orElse(null),
+        baseAmount = contract.baseSideAmount,
+        quoteAmount = contract.quoteSideAmount,
+        paymentAccountData = trade.paymentAccountData.get(),
+        bitcoinPaymentData = trade.bitcoinPaymentData.get(),
+        paymentProof = trade.paymentProof.get(),
```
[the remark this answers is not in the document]
wodoro: these are not optional, these are Observable<String>

shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/utils/DateUtils.kt
```diff
@@ -71,6 +75,27 @@ object DateUtils {
...
     }
 
+    private val mediumDateTimeFormat =
+        LocalDateTime.Format {
+            monthName(MonthNames.ENGLISH_ABBREVIATED)
+            char(' ')
+            day(Padding.NONE)
+            chars(", ")
+            year()
+            chars("  ")
```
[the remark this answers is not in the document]
wodoro: this is intentional and according to design

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/my_trades/open/OpenTradeListPresenter.kt
```diff
@@ -0,0 +1,159 @@
...
+        _uiState.update { it.copy(searchQuery = "") }
+    }
+
+    private fun onNavigateToOfferbook() {
+        navigateToTab(NavRoute.TabOfferbookMarket)
+    }
+
+    private fun onOpenTradeGuide() {
+        navigateTo(NavRoute.TradeGuideOverview)
+    }
```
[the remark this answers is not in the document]
wodoro: @rodvar  is this intentional or do we want this fix?

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
wodoro: yes.
no we are not risking, the code should cover live transition, though I have not tested it either, but thinking about it, a user on mobile cannot close a trade while on closed trades tab, so ... idk

wodoro: updated the comment

shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/usecase/trade/FilterOpenTradesUseCase.kt
wodoro: changed factory to single for this one as suggested by CRAI, let me know if that needs to be reverted (it's a frequently used screen, so I thought this should be ok)

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/mapping/trade/BisqEasyClosedTradeMapping.kt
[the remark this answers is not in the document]
wodoro: but it seems like this is an issue with bisq2 repo, I think this should be actually an optional, because it may be null? double check and create an issue on bisq2 repo if my guess is correct and there's no open issue for it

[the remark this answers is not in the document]
wodoro: bisq 2 code base rule is to specify Optional anywhere a value can be null, to have a null safe code base. this seems to be violating that.

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/my_trades/open/OpenTradeListPresenter.kt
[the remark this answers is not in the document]
rodvar: > @rodvar is this intentional or do we want this fix?

def. not - although I have never come across a related bug, open trade guide usually works might need to check how the visible flag is being used?

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
rodvar: this is the paginated response in the client facade impl

```
PaginatedResponse(
                    items = paginatedResponse.items.map(ClosedTradeListItemDto::toClosedTradeListItem),
                    page = paginatedResponse.page,
                    pageSize = paginatedResponse.pageSize,
                    totalItems = paginatedResponse.totalItems,
                    totalPages = paginatedResponse.totalPages,
                )
```

page and pageSize depende on the api response 👍 

for node it always show 1 page regardless of the pageSize = total. What I'm saying is to programatically control this (paginate in the app to avoid a huge single page for pro users.

wodoro: I don't understand how a power user is going to use that
and paginating it means wasteful filtering and sorting on every page load (scrolling down?). everything is already in the memory, so why should we try to paginate it wrongfully every time ?

rodvar: mmm maybe I'm missing some context from your answer wod - let's continue talking after I test it

What I am trying to convey is the bad experience of having a long scrollable list that could have 1000 items for a power user (typically bisq2 seller).

But from your answer above it looks like it doesn't matter it would be same based on this impl? let me test it in the device and we go from there on this one 🙏

wodoro: I understand, but it's still a long list paginated or not, and the best way to navigate that is to filter, sort and search
paginating it would only cause unnecessary filter and sort on loading next page
the only real benefit would be less pressure on composable, which is a real concern and we should maybe benchmark it, but it's a Lazy list, so I'm not sure how much of an impact it will have. it will have less memory footprint if it's paginated though

rodvar: its looking really good wod well done.

On this point above, my logic is even if filter+sort runs once for the whole list, return only the requested slice via subList after sorting -> same compute as today, but the UI gets the contract it expects, and Compose composition cost scales with visible items, not the list size. 

However, I've just read a bit more on LazyColumn and it is to Compose what the RecyclerView used to be in old Android UI -> I think we are good then thank you

rodvar (changes requested):
Image
Image

its looking great in all apps (see screens above)

there is only one thing we need to adjust UX wise to have this ready to merge

UX expected for the bottom sheet filter component is "apply on demand" for every interaction.
Wod added apply/reset buttons because the call could be expensive (tor connect for example). Upon discussion we decided to keep the component UX and solve it with coroutines (400 ms debounce + cancel and fire on each change) to tackle the scenario where the user changes several filters.

Wodoro is helping with Bisq1 at the moment - agreed with him to continue where he left off to finish this feature will do soon

rodvar:
I'll also try to improve the PR coverage (currently ~10%)

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
wodoro: The thing is if you stick to the contract exactly, you will run the filter and sort more times than actually necessary
Every user scroll that will trigger fetching next page will trigger filter and sort from scratch, which is unnecessary. The result is not changed, but you add cost to the operation
It's only beneficial if the pagination happens down to bisq2 core, otherwise I think its wasteful to sublist at the end (unless the entire list causes more memory pressure)

rodvar:
converting to draft whilst changes are being made

rodvar:
also fix the "noop functionality" when the api is not avail (this is important because when releasing this most probably Bisq2 Desktop 2.1.11 won't be released yet) - commit incoming with that first

rodvar:
rebased to main

PR coverage improved 10% -> 25%

project avg: `77.1074%` well over the target (should be updated by now :) 

Will address CRAI comments and then merge after bisq2 PR gets merged.

rodvar:
conflicts fixed and branch rebased

shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/tabs/open_trades/ClosedTradeListPresenterTest.kt
```diff
@@ -0,0 +1,226 @@
...
+            tickFlow.value = 1
+            // Before debounce fires: no invalidation yet
+            advanceTimeBy(200)
+            // After 300ms debounce
+            advanceTimeBy(200)
+            // refreshTick has now incremented — totalCount resets to null on each new QueryKey
+            // (onEach { _totalCount.value = null } runs before flatMapLatest)
+            // We can't directly inspect refreshTick, but we can verify the presenter is still alive
+            assertEquals("", localPresenter.uiState.value.searchQuery)
+
```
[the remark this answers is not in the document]
rodvar: after reviewing I think we'll just delete this test, I'll do it before merging

rodvar:
fixed conflicts locally - will push soon

rodvar:
rebased to main

rodvar (approved):
ACK

PR #1433, opened 2026-05-19
Implement Wise account (form, detail) to the musig flow, adding search feature to account dialogs

salottodev:
- contribs to #991 
 - depends on https://github.com/bisq-network/bisq2/pull/4729

## Summary by CodeRabbit

* **New Features**
  * Wise payment method added: create/manage Wise accounts with multi-currency selection, holder name and email; new review/detail and form flows.

* **Bug Fixes / UX**
  * Improved duplicate-account handling: API conflict maps to a friendly validation message and shows an error snackbar.

* **UI**
  * Currency picker: search, select-all, clear-all, multi-select summary and formatted display; account dialogs support highlighted search.

* **Tests**
  * Added unit and UI tests covering Wise form, review, mapping, and account flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bisq-network/bisq-mobile/pull/1433?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1435, opened 2026-05-20
Make musig payment flow only available in client module

salottodev:
Depends on: https://github.com/bisq-network/bisq-mobile/pull/1433


rodvar:
@coderabbitai review - focus on the 2nd commit as the first one is part of a previous PR dependency

salottodev:
@coderabbitai
Regarding to your actionable nits and suggestions:

1-  The suggested import is not correct. ConsistentCopyVisibility comes from the Kotlin stdlib package 

2-  I don’t think this route currently has a deep-link exposure: ClientNavRoute.PaymentAccountsMusig is only registered as a typed Compose destination and does not implement DeepLinkableRoute or define a navDeepLink in ClientRootNavGraph. The menu entry is gated by MU_SIG_ENABLED, so the normal user path is covered.

3- Tests are running well and passing.

4- I will apply the fix.

5- This is optional and not needed.

6- "common" path is the right path

rodvar:
conflicts because the first commit is dependent on a merged commit with a different hash, simply remove that one and rebase the branch and should be ok 👍

rodvar (approved):
utACK 

in-between refactor, will be tested along with next payment methods PRs

rodvar:
PR coverage check ignored for this refactor PR

PR #1453, opened 2026-06-02
Implement Revolut payment account flow

salottodev:
## Release Notes

* **New Features**
  * Added support for Revolut payment accounts with username and multi-currency selection.
  * Introduced reusable currency picker component with search functionality for improved usability.

* **Bug Fixes**
  * Updated currency picker labels to use consistent, generic naming across payment methods.

* **Tests**
  * Added comprehensive test coverage for Revolut account creation, review, and detail screens.


rodvar:
@coderabbitai full review

rodvar:
@coderabbitai I don't see the results of your full review above, what happened?

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/core/ui/CurrencyPickerBottomSheet.kt
```diff
@@ -0,0 +1,207 @@
...
+            BisqText.SmallLight(
+                text = "mobile.paymentAccounts.currencyPicker.editLabel".i18n(),
+                color = if (isError) BisqTheme.colors.danger else BisqTheme.colors.mid_grey20,
+            )
+        }
+    }
+}
+
+@Composable
+fun CurrencyPickerBottomSheet(
```
rodvar, line 69: 👏

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/mapping/account/fiat/FiatCurrencyMapping.kt
```diff
@@ -0,0 +1,6 @@
+package network.bisq.mobile.data.mapping.account.fiat
+
+import network.bisq.mobile.data.model.account.fiat.FiatCurrencyDto
+import network.bisq.mobile.domain.model.account.fiat.FiatCurrency
+
+internal fun FiatCurrencyDto.toDomain(): FiatCurrency = FiatCurrency(code = code, name = name)
```
rodvar, line 6: 👍

shared/domain/src/commonMain/resources/mobile/account.properties
```diff
@@ -239,15 +234,15 @@ paymentAccounts.createAccount.accountData.backgroundOverlay.zelle=Zelle is a mon
...
-mobile.paymentAccounts.wise.currencies.summary={0} of {1} currencies selected
-mobile.paymentAccounts.wise.currencies.allSelected=All {0} currencies selected
-mobile.paymentAccounts.wise.currencies.editLabel=Tap to change
-mobile.paymentAccounts.wise.currencies.error=Select at least one currency
-mobile.paymentAccounts.wise.picker.title=Supported currencies
-mobile.paymentAccounts.wise.picker.selectAll=Select all
-mobile.paymentAccounts.wise.picker.clearAll=Clear all
-mobile.paymentAccounts.wise.picker.searchHint=Search currencies
-mobile.paymentAccounts.wise.picker.noResults=No currencies match your search
+mobile.paymentAccounts.currencyPicker.summary={0} of {1} currencies selected
```
rodvar, line 237: 👏

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/payment_account_form/form/revolut/RevolutFormPresenter.kt
```diff
@@ -0,0 +1,172 @@
...
+                        currencyErrorMessage = null,
+                    )
+                }
+            }
+
+            RevolutFormUiAction.OnClearAllCurrencies -> {
+                _uiState.update { it.copy(selectedCurrencyCodes = emptySet()) }
+            }
+
+            else -> Unit
```
rodvar, line 105: intentional?

shared/domain/src/commonMain/resources/mobile/account.properties
rodvar: let's keep an eye on transifex today to see if it cleans up the now orphaned paymentAccounts.revolut.* keys

rodvar (commented):
first code review pass looking good, will test soon and put further comments 💪

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/payment_account_form/form/revolut/RevolutFormPresenter.kt
salottodev: yes, but this will be improved in a future PR, so we can safely remove that else.

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/payment_account_form/form/revolut/RevolutFormPresenter.kt
```diff
@@ -0,0 +1,172 @@
...
+open class RevolutFormPresenter(
+    mainPresenter: MainPresenter,
+) : AccountFormPresenter(mainPresenter) {
+    private val _uiState = MutableStateFlow(RevolutFormUiState())
+    override val uiState: StateFlow<RevolutFormUiState> = _uiState.asStateFlow()
+
+    private val _effect = MutableSharedFlow<RevolutFormEffect>()
+    val effect = _effect.asSharedFlow()
+
+    private var supportedCurrencies: HashSet<FiatCurrency> = HashSet()
```
rodvar, line 33: why this is not part of the uiState? (I think I missed this in Wise too)

salottodev: I keep that in presenter memory to have O(1) access when doing some calculations. Please check the usage.

salottodev: It is also part of uiState like this:

```
 availableCurrencies =
                    supportedCurrencies.sortedBy { currency -> currency.code }.map { currency ->
                        CurrencyPickerItem(
                            code = currency.code,
                            displayName = currency.toDisplayString(),
                        )
                    },
```

It is mapped to the object that is used to display the bottom picker.

rodvar: I had a quick check, I don't think this is performance related - can you be more specific on how O(1) access is being used here?

The code you posted as well as `supportedCurrencies.filter { it.code in selectedCurrencyCodes }.sortedBy { it.code }` are both O(n) -> it will iterate all items regardless of structure being list or (hash)set.

Maybe you went this way because of complications with the CurrencyPicketItem display name? Maybe consider either:

 1) add `supportedCurrencies: Set<FiatCurrency>` to the UiState or
 2) add `name: String` to the CurrencyPickerItem (I think this is the best and will benefit wise as well :) )

this won't prevent PR merging but good to discuss for a follow up improvement

rodvar (approved):
ACK

2 extra comments:

 - please check my last code review comment for a follow up / next PR
 - see screenshot below: Revolut users use the "revtag" (something like @myrevtag) to easily find each other but our wording is `username`. I think is not a big deal since there are other ways of finding a revolut user we can leave that as a free text and they can still use the "@" to signal to each other. just worth mentioning

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/payment_account_form/form/revolut/RevolutFormPresenter.kt
salottodev: This where I wanted to get O(1) access 

```
    override fun onNextClick() {
        var selectedCurrencies = emptyList<FiatCurrency>()
        val validatedState =
            _uiState.updateAndGet {
                selectedCurrencies =
                    supportedCurrencies
                        .filter { currency -> it.selectedCurrencyCodes.contains(currency.code) }
                        .sortedBy { currency -> currency.code }
  ```
  
  This line:
  
  ```
  it.selectedCurrencyCodes.contains(currency.code) }

**Comment by salottodev** [comment]
@rodvar regarding to Revolut username label we are using the same label that desktop is using. We should match both platforms. If we want to update, I think we should talk to Henrik

File: apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/payment_account_form/form/revolut/RevolutFormPresenter.kt (hunk 5 above)
Comment (rodvar) [reply apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/create_payment_account/payment_account_form/form/revolut/RevolutFormPresenter.kt:33]: correct: that line is O(1) **but** the discussion is not with `selectedCurrencyCodes` (which is already part of uiState), its with the presenter supportedCurrencies which is iterated O(n) :

```
supportedCurrencies
                        .filter { currency -> it.selectedCurrencyCodes.contains(currency.code) }
```

Shouldn't we respect the MVIP pattern and move that one into the uiState?

rodvar: upon chat and double-checking I see that supportedCurrencies is not used for UI reaction 👍 

I do think though my argument I was trying to convey above `supportedCurrencies exists only because CurrencyPickerItem drops the name field that onNextClick needs to rebuild FiatCurrency` still holds and one of the 2 options provided would eliminate the need entirely, but I leave it for you to decide on further dev! thank you

PR #1454, opened 2026-06-02
bugfix/opted_in_android_relayed_notifications_too_many_per_open_trade

### LINKED ISSUE #1450: Notifications: opted-in relayed notifications send to many updates per active trade
probably we missed some guards already implemented for the nodeApp case that are needed as well in the api-based relayed solution or could be an FCM specific issue.

rodvar:
- resolves #1450 
 - depends on https://github.com/bisq-network/bisq2/pull/4797
 - avoid FG service reposting already managed by relayed mechanism notifications
 - fix regression bug preventing foreground service to start when keep alive is selected
 - fix bug showing a chat message generic message as a trade message
 - improve docs + tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification categorization so chat messages are no longer miscategorized as trade updates.
  * More consistent background behavior: foreground keep-alive and local notification suppression are now handled separately, reducing missed or duplicate notifications and improving connection persistence when enabled.

* **Tests**
  * Added coverage for chat-vs-trade notification precedence and for the new foreground/suppression lifecycle behaviors, including backward-compatibility scenarios.

### DIFF
--- a/apps/clientApp/src/androidMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/BisqFirebaseMessagingService.kt
+++ b/apps/clientApp/src/androidMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/BisqFirebaseMessagingService.kt
@@ -318,9 +318,19 @@ class BisqFirebaseMessagingService :
 
             internal fun fromTitle(title: String): NotificationCategory {
                 val lower = title.lowercase()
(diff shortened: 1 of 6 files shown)


rodvar:
@coderabbitai full review

rodvar:
bisq2 maintainer not available. Dependency on bisq2 PR is a soft one (each fixes a different issue that contributes to the same user-facing issue), so will improve coverage and merge this one to include in releases

PR #1461, opened 2026-06-04
Add bank data details by country endpoint

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added retrieval of country-specific bank account requirements and details
  * Implemented caching mechanism for improved performance on repeated requests

* **Tests**
  * Extended test coverage for bank account country details retrieval and caching behavior

### DIFF
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/repository/BankAccountCountryDetailsRepositoryImpl.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/repository/BankAccountCountryDetailsRepositoryImpl.kt
@@ -1,0 +1,74 @@
+package network.bisq.mobile.data.repository
+
+import androidx.datastore.core.DataStore
+import androidx.datastore.core.IOException
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import network.bisq.mobile.client.shared.BuildConfig
+import network.bisq.mobile.data.model.account.bank_account.BankAccountCountryDetailsCache
+import network.bisq.mobile.data.model.account.fiat.BankAccountCountryDetailsDto
+import network.bisq.mobile.domain.repository.BankAccountCountryDetailsRepository
+import network.bisq.mobile.domain.utils.Logging
+
+class BankAccountCountryDetailsRepositoryImpl(
+    private val cacheStore: DataStore<BankAccountCountryDetailsCache>,
+) : BankAccountCountryDetailsRepository,
+    Logging {
+    private val mutex = Mutex()
+    private var detailsByCountryCode: Map<String, BankAccountCountryDetailsDto> = emptyMap()
+    private var loadedApiVersion: String? = null
+
+    override suspend fun get(
+        countryCode: String,
+        refresh: suspend () -> List<BankAccountCountryDetailsDto>,
+    ): BankAccountCountryDetailsDto {
+        val normalizedCountryCode = normalizeCountryCode(countryCode)
+        require(normalizedCountryCode.isNotBlank()) { "countryCode cannot be blank" }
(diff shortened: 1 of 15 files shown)


shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/repository/BankAccountCountryDetailsRepositoryImpl.kt
```diff
@@ -0,0 +1,74 @@
...
+        countryCode: String,
+        refresh: suspend () -> List<BankAccountCountryDetailsDto>,
+    ): BankAccountCountryDetailsDto {
+        val normalizedCountryCode = normalizeCountryCode(countryCode)
+        require(normalizedCountryCode.isNotBlank()) { "countryCode cannot be blank" }
+
+        return mutex.withLock {
+            ensureLoaded(refresh)
+            detailsByCountryCode[normalizedCountryCode]
+                ?: error("Bank account country details not found for country code: $normalizedCountryCode")
```
rodvar, line 32: better to use a specific exception like NoSuchElementException, or a project-specific `CountrySupportException` ?
Otherwise presentation doesn't know why its failing to present to the user accordingly?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/repository/BankAccountCountryDetailsRepositoryImpl.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val mutex = Mutex()
+    private var detailsByCountryCode: Map<String, BankAccountCountryDetailsDto> = emptyMap()
+    private var loadedApiVersion: String? = null
+
+    override suspend fun get(
+        countryCode: String,
+        refresh: suspend () -> List<BankAccountCountryDetailsDto>,
+    ): BankAccountCountryDetailsDto {
+        val normalizedCountryCode = normalizeCountryCode(countryCode)
+        require(normalizedCountryCode.isNotBlank()) { "countryCode cannot be blank" }
```
rodvar, line 27: please add unit test for this case, same for the IOException catch in readCache()

rodvar (changes requested):
Very solid implementation 👏 

Justa couple of things to check and we can merge this one.

And one question `BankAccountCountryDetailsCache.kt` lives in `bank_account` sub package, but the rest of the impl is in `fiat` subpackage.. is that intentional?

salottodev:
> Very solid implementation 👏
> 
> Justa couple of things to check and we can merge this one.
> 
> And one question `BankAccountCountryDetailsCache.kt` lives in `bank_account` sub package, but the rest of the impl is in `fiat` subpackage.. is that intentional?

Regarding to this, a PR with a package re organization is coming.

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/repository/BankAccountCountryDetailsRepositoryImpl.kt
salottodev: As the user has to select the country from a dropdown and it is not something that user writes, the case where the country is not found is very edge case, that's why I decided not to handle it. If we really want to handle this scenario we should provide a i18n error message. I don't think this is necessary, but let me know if I should implement this or not.

rodvar (approved):
utACK

will be tested with next PR + flagged issues will be addressed as discussed with salotto

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/repository/BankAccountCountryDetailsRepositoryImpl.kt
salottodev: Ok, I will add in a future PR

PR #1470, opened 2026-06-08
Move account models, mappings and tests to client module

salottodev:


rodvar (approved):
utACK

refactor packaging restructure

PR #1485, opened 2026-06-12
Implement bank payment account to musig flow

salottodev:
## Release Notes

* **New Features**
  * Added support for Same Bank and National Bank payment account types
  * Implemented forms for creating and managing these new account types
  * Added detailed account views with country-specific field validation
  * Integrated new payment methods into account review and management flows

### DIFF
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/payment_accounts/presentation/payment_accounts_list/model/FiatAccountVO.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/payment_accounts/presentation/payment_accounts_list/model/FiatAccountVO.kt
@@ -7,6 +7,7 @@ import network.bisq.mobile.client.payment_accounts.domain.model.fiat.ach_transfe
 import network.bisq.mobile.client.payment_accounts.domain.model.fiat.cash_deposit.CashDepositAccount
 import network.bisq.mobile.client.payment_accounts.domain.model.fiat.common.country.FiatPaymentCountryBasedAccountPayload
 import network.bisq.mobile.client.payment_accounts.domain.model.fiat.common.currency.FiatPaymentSingleCurrencyAccountPayload
+import network.bisq.mobile.client.payment_accounts.domain.model.fiat.national_bank.NationalBankAccount
 import network.bisq.mobile.client.payment_accounts.domain.model.fiat.revolut.RevolutAccount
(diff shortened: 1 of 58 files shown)


rodvar:
https://github.com/bisq-network/bisq2/pull/4812 already merged

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/payment_accounts/presentation/payment_accounts_list/model/FiatAccountVO.kt
```diff
@@ -34,6 +35,7 @@ fun FiatPaymentAccount.toVO(): FiatAccountVO? {
             is ZelleAccount -> PaymentTypeVO.ZELLE
             is WiseAccount -> PaymentTypeVO.WISE
             is RevolutAccount -> PaymentTypeVO.REVOLUT
+            is NationalBankAccount -> PaymentTypeVO.NATIONAL_BANK
             is SepaAccount -> PaymentTypeVO.SEPA
```
rodvar, line 39: shouldn't have `is SameBankAccount -> PaymentTypeVO.SAME_BANK` here as well?

rodvar (approved):
ACK

Overall - great work

I put the issues I found in order of importance here. Normally, number 1 would be a blocker for merge but since this feature is fully disabled it can also be done in a follow up.

  1. add is SameBankAccount -> PaymentTypeVO.SAME_BANK in FiatAccountVO.toVO()
  2. From a UX perspective, it would be good that changing the country combo automatically selects that country national currency (see screenshot below), and the user then can decided if he edits it further (enhancement, we probably want to create an issue for this one and inplement in bisq2?)
  3. Same Bank and National Bank DTOs are almost identical with all the overhead of the tests and stuff that can be simplified into one class + extension if necessary? consider refactor there
  4. Nice: extend SameBankAccountDetailPresenterTest to also cover NationalBankAccount (same presenter, free coverage)
  5. add a log.e in the PaymentAccountForm screen's else branch for the nullable-cast silent-fail path (consistency with the rest of the codebase)

Image

rodvar:
@salottodev sorry for the lack of precision in the last review comment (5), I was referring to the else that uses `UnsupportedAccountState`

this is a pattern that happens across the payment methdos impl, maybe you can think about a way of generalising it and add the log so devs know if that happens 🙏

PR #1486, opened 2026-06-12
Feature/opt in analtics wrap up and dashboard promo

### LINKED ISSUE #525: Opt-in private crash and performance dev-friendly reporting
## Requirements:
 
- disabled by default, user should be prompted to opt in only once and then referred to settings if he wants to opt in/out) 
 - privacy first (no surveillance)
 - open sourced
 - if centralised, server/s managed by us
 - time based auto deletion of user data

### Implementing this has the following benefits

 - It works regardless of the build distribution method the user used to get the app/s
 - We can track much more than just the crash (performance, time metrics)
 - accuracy: crash reports will tell you exact class and line related to crash + excellent device data
 - without setting this we will only have vague anonymous data for the users using Google Play. Better than having 0, but more difficult to work bugs and crashes of specific devices.

A strong candidate to implement this is [Sentry](https://sentry.io/welcome/) being open sourced and having a standalone option to self host.
We could use our testing server to set it up.

rodvar:
Screenshot 2026-06-12 at 10 29 28

 - resolves #525 
 - add carrousel promo to encourage people to opt in (dismiss or opt in options)
 - adds basic main screens views analytics + settings + once off baseline
 - substantial docs + tests for all 3

## Summary by CodeRabbit

* **New Features**
  * Emits an analytics settings baseline once per opt-in (captures analytics opt-in, normalized language, push, and keep-connected).
  * Adds an analytics card to the welcome carousel with “Help improve Bisq Mobile” and a learn-more CTA.
  * Expands screen analytics coverage using explicit “screen opened” events and adds language-change analytics.

* **Bug Fixes**
  * Fixes analytics language handling to avoid crashes by safely normalizing nullable/blank/unrecognized language codes.

* **Tests**
  * Adds/updates unit, contract, and analytics coverage tests for baseline emission, event naming, screen coverage, and language observer behavior.


shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/analytics/AnalyticsSettingsBaseline.kt
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+        // Push notifications opt-in
+        analyticsService.track(
+            if (pushNotificationServiceFacade.isPushNotificationsEnabled.value) {
+                AnalyticsEvent.Settings.PushNotificationsEnabled
+            } else {
+                AnalyticsEvent.Settings.PushNotificationsDisabled
+            },
+        )
```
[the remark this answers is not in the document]
rodvar: it is using the right persisted value iin the stateflow

shared/domain/src/commonMain/kotlin/network/bisq/mobile/domain/analytics/AnalyticsEvent.kt
```diff
@@ -17,27 +17,246 @@ package network.bisq.mobile.domain.analytics
...
+                    CreateOfferReview,
+                    TakeOfferAmount,
+                    TakeOfferPaymentMethod,
+                    TakeOfferReview,
+                )
+            }
+        }
+
+        // -- Tier A: core funnel spine ---------------------------------
+        data object Splash : ScreenViewed("screen.splash_opened")
```
salottodev: May be better `screen.xxx_viewed` to sync with `ScreenViewed`?

rodvar: thanks for having a look into this!

Do you think is worth to change all of the `opened` to `viewed` ? No strong opinion but can do before merging this

rodvar: we discussed and agreed to change the object name instead to `ScreenOpened` - will do now before merge

rodvar:
@coderabbitai this PR is making the anayltics service nullable and using lazy loading techniques to avoid crashes in tests and doing the proper injection of the mock service that would imply editing several classes.

Can we create a follow up issue to improve this in the future? please feel free to review the related code and add your comments on it.

PR #1523, opened 2026-06-19
Replace global disableInteractive() with per-action guards and guardedSuspendAction (#1284)

### LINKED ISSUE #1284: Add double-tap/duplicate-call protection to unguarded presenter methods
## Context

As part of the migration away from the global `isInteractive` guard pattern (tracked in #1245, introduced in PR #1244), the following presenter methods have been identified as lacking protection against duplicate/concurrent calls (e.g. rapid double-taps).

The new pattern to apply is: a local `isLoading`/`showLoading()` guard at the start of each method, replacing any remaining `disableInteractive()` / `enableInteractive()` usage.

---

## 🔴 CRITICAL — Trade protocol actions (financial impact)

These all currently follow the pattern `presenterScope.launch { showLoading(); service.call() }` **with no duplicate-call guard**:

1. `SellerState1Presenter.onSendPaymentData()` — sends payment account data
2. `BuyerState1aPresenter.sendBitcoinPaymentData()` — sends BTC address
3. `BuyerState2aPresenter.onConfirmFiatSent()` — confirms fiat sent
4. `SellerState2bPresenter.onConfirmFiatReceipt()` — confirms fiat received
5. `SellerState3aPresenter.confirmSend()` — confirms BTC sent
6. `BuyerStateLightning3bPresenter.onCompleteTrade()` — completes lightning trade
7. `SellerStateLightning3bPresenter.skipWaiting()` — skips waiting, confirms BTC
8. `State4Presenter.onConfirmCloseTrade()` — closes trade

---

## 🟠 HIGH — Account management

9. `PaymentAccountsPresenter.addAccount()` — creates account
10. `PaymentAccountsPresenter.saveAccount()` — saves account
11. `PaymentAccountsPresenter.deleteSelectedAccount()` — deletes account

---

## 🟡 MEDIUM — Settings & chat (low risk, mostly idempotent)

12. `SettingsPresenter` — multiple save methods (language, trade price tolerance, etc.)
13. `TradeChatPresenter.sendChatMessage()` — could send duplicate messages
14. `ReportUserPresenter.onReportClick()` — report user action

---

## ✅ Already protected (need pattern replacement only)

These use `disableInteractive()` and should be migrated to the local `showLoading()`/`hideLoading()` pattern:

- `TakeOfferReviewPresenter.onTakeOffer()`
- `CreateProfilePresenter.onCreateAndPublishNewUserProfile()`
- `UserProfilePresenter.onSavePress()` / `onDeleteConfirm()`

---

## Tasks

- [ ] Add `if (_isLoading.value) return` (or equivalent `showLoading()`-based guard) to each unprotected method above
- [ ] Prioritise CRITICAL methods first
- [ ] Migrate already-protected methods from `disableInteractive()` to `showLoading()`/`hideLoading()`
- [ ] Ensure UI buttons are disabled while loading state is active

## References

- Pattern introduced in: https://github.com/bisq-network/bisq-mobile/pull/1244
- General migration tracking: #1245
- PR comment with analysis: https://github.com/bisq-network/bisq-mobile/pull/1244#issuecomment-2738440015

Requested by: @rodvar

nostrbuddha:
Fixes https://github.com/bisq-network/bisq-mobile/issues/1284

Replaces the global `enableInteractive() / disableInteractive() / isInteractive` pattern with `showLoading() / hideLoading()`, local per-action guards on presenters, by using `BasePresenter.guardedSuspendAction`.

Removed isInteractive and associated code from BasePresenter and all presenters in the repo.

## Summary by CodeRabbit

* **New Features**
  * Retry/confirm buttons and other action buttons now show loading and disable based on action readiness, improving responsiveness during multi-step flows.

* **Bug Fixes**
  * Prevented duplicate submissions by guarding actions against rapid repeated taps; buttons now re-enable appropriately after failures.

* **Refactor**
  * Replaced broad “freeze UI” behavior with action-specific enablement and unified guarded handling across key screens (offers, trades, settings, payment accounts, onboarding, dialogs).

* **Tests**
  * Expanded automated coverage for guarded actions, guard reset, and duplicate-call prevention.


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/offerbook/OfferbookPresenter.kt
```diff
@@ -120,6 +129,7 @@ open class OfferbookPresenter(
     override fun onViewAttached() {
         super.onViewAttached()
 
+        resetActionGuards()
         selectedOffer = null
```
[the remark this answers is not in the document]
nostrbuddha: A call to onViewAttached(), while delete offer is in process is not possible with the way the UI is setup now. We can safely ignore this.

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/report_user/ReportUserPresenter.kt
```diff
@@ -38,39 +40,44 @@ class ReportUserPresenter(
...
             _uiState.update { it.copy(isLoading = true) }
-            userProfileServiceFacade
-                .reportUserProfile(
-                    chatMessage.senderUserProfile,
-                    message,
-                ).onSuccess {
-                    _effect.emit(ReportUserEffect.ReportSuccess)
-                }.onFailure {
+            val message = _uiState.value.message
+            try {
```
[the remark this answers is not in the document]
nostrbuddha: This wont be needed. guardedSuspendAction() locks the entire screen from user interaction, so message cannot be changed once it kicks in.

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/settings/SettingsPresenter.kt
```diff
@@ -349,23 +374,25 @@ open class SettingsPresenter(
...
+        val newDeviation = parsedValue / 100
+
+        guardedSuspendAction(_isTradePriceToleranceSaveEnabled, "onTradePriceToleranceSave") {
+            settingsServiceFacade
+                .setMaxTradePriceDeviation(newDeviation)
+                .onSuccess {
                     originalMaxTradePriceDeviation = newDeviation
                     _uiState.update {
                         it.copy(hasChangesTradePriceTolerance = false)
                     }
```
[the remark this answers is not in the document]
nostrbuddha: This wont be needed. guardedSuspendAction() locks the entire screen from user interaction, so user can't interact with any of the screen elements, when guardedSuspendAction() is happening

rodvar:
@coderabbitai full review

rodvar (approved):
ACK 

very solid PR to finish the job

I think we should apply the same logic for the hideLoading as well -> change to scheduleHideLoading() instead and add the short 150ms delay before clearing up the job and flags.

It's noticeable is needed for example in the onboarding when the create/profile is needed (new install of node, or first user for connect trusted node) there is a very brief moment where the user could tap on stuff before the navigation happens.

If you agree please send a follow up PR 🙏  

cc: @salottodev

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/base/BasePresenter.kt
```diff
@@ -526,6 +477,41 @@ abstract class BasePresenter(
         globalUiManager.hideLoading()
     }
 
+    /**
+     * Runs [block] on [presenterScope] behind [guard] using compareAndSet.
+     * [guard] uses `true` = enabled; it is cleared for the duration of [block]
+     * and restored in finally when [reEnableGuardOnComplete] is true.
+     * Returns whether the action was started.
+     */
+    protected fun guardedSuspendAction(
```
salottodev, line 486: @nostrbuddha @rodvar 

I wonder if guardedSuspendAction is taking on a bit too much responsibility here. From the name/API, I’d expect it to only guard against duplicate concurrent execution, but it also controls global loading UI via showLoadingOverlay.

That makes the helper harder to reason about: callers that only want guarded execution may accidentally get loading behavior, and callers that need custom/local loading have to opt out. I think loading would be clearer if handled explicitly at the call site.

Also, it looks like none of the production callers are using the returned Boolean. If this helper is going to return something, would returning a Job? be more useful? The current Boolean only tells us whether the action was started, but it doesn’t give the caller any control over the launched coroutine. If a caller ever needs to cancel the work manually, there’s no handle to do that. Returning null when the guard rejects the action and a Job when it starts would preserve the “started vs ignored” signal while also giving callers a cancellation handle.

rodvar: Thanks @salottodev, both fair points.

On the return type: agreed Job? is the stronger signature — null = rejected, Job = started + a cancel handle, strictly more thanthe Boolean. Only reason I'd hold off is that no caller needs the handle yet, so I'd rather flip it to Job? the day one actually, vs. adding optionality we don't use. I think this is a good follow up to be done if you would like to take it?

On the loading coupling: that one's intentional. Almost every guarded action here is a network call that wants the overlay, so bundling it (default-on, opt-out flag) is what keeps the ~50 call sites boilerplate-free - splitting it back to each call site trades a mild code smell for more boilerplate everywhere, which is the thing this helper was meant to remove (this was originally discussed and personally requested to buddha)  

So I'd lean: keep as-is for now, and open a small follow-up for the Job? switch if/when a caller needs cancellation. Sound good? 💪

nostrbuddha: Yes. Returning the Job will be a more useful, than the current bool return value.

rodvar: @coderabbitai please read my comment above and let's create an issue follow up please

PR #1535, opened 2026-06-23
fix(nodeApp): persist trade messages after Bisq2 rate-limited channel write

### LINKED ISSUE #1421: MyTrades :: Notification badge count issue
Steps: 
 - Say you have 2 unread notifications for a single trade.
 - Close the mobile app.
 - Sent 3 more messages from your peer.
 - Open the mobile app.

Expected:
 - 5 unread notifications in the 3rd tab.

Observed:
 - 4 unread notifications in the 3rd tab.

Happens only in androidNode. androidClient dont have this issue

nostrbuddha:
- Fixes https://github.com/bisq-network/bisq-mobile/issues/1421

Easy steps to reproduce:
1. Take an offer (Observe 1 notification for this newly taken offer)
2. Close the app.
3. From the trade peer, send payment account data and 1 message
4. Not open the app.

Expected:
3 notification (1 for Offer taken, one for payment account data, one for chat message)

Observed:
2 notification (One for payment account data, one for chat message)
Offer taken message and respective notification is not shown.

Root cause:
On taking an offer, androidNode receives `TAKE_BISQ_EASY_OFFER` and ~200 ms later the `PROTOCOL_LOG_MESSAGE`.“ This first PROTOCOL_LOG_MESSAGE maps to the system message: "X has sent a message for taking Y's offer”.
Same scenario plays out, even when a peer takes your offer.

Bisq2's `RateLimitedPersistenceClient` drops the second `persist()` since it falls inside the 1 s window. The log appears live in memory but never reaches disk; after an app kill, without any other user interection that might trigger persist(), this message is gone.

So on next app re-start, it's not shown.

We had similar persist issue when working with Settings and did a similar persist logic @ NodeSettingsServiceFacade::persistWithRetry()

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat message handling so recent activity and history load more reliably.
  * Prevented certain message types from appearing in chat and ensured protocol log messages are saved more consistently.
  * Updated coroutine test support across the app to keep automated coverage aligned with the latest shared test utilities.

* **Tests**
  * Added and updated tests covering chat message loading, persistence timing, and trade-state edge cases.


rodvar:
rebased to main

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/chat/trade/NodeTradeChatMessagesServiceFacade.kt
```diff
@@ -127,49 +136,17 @@ class NodeTradeChatMessagesServiceFacade(
...
-                                                    .map { reaction ->
-                                                        Mappings.BisqEasyOpenTradeMessageReactionMapping.fromBisq2Model(
-                                                            reaction,
-                                                        )
-                                                    }
-                                            model.setReactions(chatMessageReactions)
-                                        }
-                                }
-                            reactionsPinByMessageId[messageId] = pin
+                    override fun onAllAdded(values: Collection<out BisqEasyOpenTradeMessage>) {
```
rodvar, line 141: The `onAllAdded == startup replay` assumption is correct **today**, but it's a fragile coupling to Bisq2 internals. I think the only reason `persist = false` is safe here is that trade messages happen to be added one-by-one — `ChatChannel.addChatMessage` -> `getChatMessages().add()` -> `onAdded`.

So the day Bisq2 batch-adds or `setAll`s trade messages live (queued backlog drain, resync, ...), those live messages silently take the `persist = false` path and **the exact bug we're fixing comes back — with no test catching it**.

I think we have two options but open to your ideas:

- minimal: drop a comment here spelling out the invariant we're relying on (live trade messages only ever arrive via `onAdded`; `onAllAdded` = the subscribe-time snapshot).
- or, more robust: stop keying off the observer method. Use a per-channel `initialized` flag — first `onAllAdded` flips it, and anything added after that (whether `onAdded` or `onAllAdded`) is treated as live -> persist. Immune to upstream batching changes.

Not a blocker but just flagging for follow ups

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/chat/trade/NodeTradeChatMessagesServiceFacade.kt
```diff
@@ -182,6 +159,73 @@ class NodeTradeChatMessagesServiceFacade(
...
+                message,
+                citationAuthorUserProfile,
+                myUserProfile,
+            )
+        openTradeItem.bisqEasyOpenTradeChannelModel.addChatMessages(model)
+
+        // PROTOCOL_LOG_MESSAGE's persist() is always rate-limited (arrives ~200 ms after
+        // TAKE_BISQ_EASY_OFFER, inside the 1 000 ms window). Force a persist once the window clears.
+        if (persist && message.chatMessageType == ChatMessageType.PROTOCOL_LOG_MESSAGE) {
+            serviceScope.launch {
```
rodvar: please fix: `persist()` does a synchronous `getPersistableStore().getClone()` before the async write, and in prod `serviceScope` runs on `Dispatchers.Main.immediate` (`DefaultCoroutineJobsManager`) — so the clone lands on the main thread. Every other method in this class already hops with `withContext(Dispatchers.Default)`. Let's keep it consistent: `serviceScope.launch(Dispatchers.Default) { ... }`?

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/chat/trade/NodeTradeChatMessagesServiceFacade.kt
```diff
@@ -182,6 +159,73 @@ class NodeTradeChatMessagesServiceFacade(
...
+                delay(PERSIST_DELAY_AFTER_PROTOCOL_LOG_MS)
+                bisqEasyOpenTradeChannelService.persist()
```
rodvar, line 226: nit (optional): each live `PROTOCOL_LOG_MESSAGE` spawns its own delayed coroutine + `persist()` (the test asserts 2 persists for 2 messages). Harmless since the rate-limiter drops the extras, but a single debounced flush would be tidier. Not worth blocking on.

apps/nodeApp/src/androidUnitTest/kotlin/network/bisq/mobile/node/test_utils/TestCoroutineJobsManager.kt
```diff
@@ -0,0 +1,23 @@
+package network.bisq.mobile.node.test_utils
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import network.bisq.mobile.domain.utils.CoroutineJobsManager
+
+class TestCoroutineJobsManager(
```
salottodev: Hey @nostrbuddha, we already have `TestCoroutineJobsManager` in another module. Please consider moving it to `test-utils` module, and check if we can reuse.

rodvar (changes requested):
Image

Its looking good overall but there is a blocker issue besides the rest of my comments, the most important one is the persist in the main thread it can (and I repro it) cause an ANR in node app.

I added some comments I think we should tackle in a follow up but are not blocker to this PR, please review 🙏 

Also, I think this leaves a gap on connect apps where this issue, although with low probabilty, could still happen. We would need a fix on bisq2 api module, shall I create an issue for this unless you have a clear idea and want to lodge it quick? 

Looking forward to the next (final) review 🙏 

thank you buddha

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/chat/trade/NodeTradeChatMessagesServiceFacade.kt
nostrbuddha: For now added a comment, mentioning this logic depends on bisq2 contract

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/chat/trade/NodeTradeChatMessagesServiceFacade.kt
nostrbuddha: Thanks for catching this. Updated to use 'serviceScope.launch(Dispatchers.Default)'

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/chat/trade/NodeTradeChatMessagesServiceFacade.kt
nostrbuddha: Assuming a trader might have a max of 10 trades open at a time, and even all those 10 received 'PROTOCOL_LOG_MESSAGE', the impact wont be much.
As you said, this is harmless. I think we can leave it as it is.

apps/nodeApp/src/androidUnitTest/kotlin/network/bisq/mobile/node/test_utils/TestCoroutineJobsManager.kt
nostrbuddha: Good one @salottodev. Doing it

rodvar (approved):
ACK (node)

all concerns addressed 💪 

there is a minor cleanup to be done from the TestCoroutine..Job I'll add in a commit please review

PR #1541, opened 2026-06-25
Add FAQs sreen implementation to More tab

salottodev:
## Summary by CodeRabbit

* **New Features**
  * Added a new FAQs screen with expandable questions and a “want to know more” link that opens the frequently asked questions page.
  * Included FAQs in the More menu and added dedicated navigation/route to the FAQs screen.
  * Added FAQ localization content and wired separate FAQ presenters for the mobile client and node views.
* **Tests**
  * Added unit and Compose UI tests covering FAQ rendering, expand/collapse behavior, and the “want to know more” action.

### DIFF
--- a/shared/domain/src/commonMain/resources/mobile/mobile.properties
+++ b/shared/domain/src/commonMain/resources/mobile/mobile.properties
@@ -274,6 +274,7 @@ mobile.user.paymentAccounts.unsupported=This account is not supported
 # More
 ################################################################################
 mobile.more.support=Support
+mobile.more.faqs=FAQs
 mobile.more.reputation=Reputation
 mobile.more.userProfile=User profile
 mobile.more.paymentAccounts=Payment accounts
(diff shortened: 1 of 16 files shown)


rodvar:
@coderabbitai full review

shared/domain/src/commonMain/resources/mobile/mobile.properties
```diff
@@ -343,6 +344,29 @@ mobile.resources.backup.success=Backup created successfully
 
 mobile.resources.restore.error=An error occurred while restoring the backup: {0}
 mobile.resources.restore.error.cannotReadFile=Could not read backup file
+
```
rodvar, line 354: nit- FAQs section placement breaks `mobile.resources.restore.error.*` in the middle

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/faqs/FaqScreen.kt
```diff
@@ -0,0 +1,255 @@
...
+import network.bisq.mobile.presentation.common.ui.components.molecules.TopBar
+import network.bisq.mobile.presentation.common.ui.components.molecules.TopBarContent
+import network.bisq.mobile.presentation.common.ui.theme.BisqTheme
+import network.bisq.mobile.presentation.common.ui.theme.BisqUIConstants
+import network.bisq.mobile.presentation.common.ui.utils.ExcludeFromCoverage
+import network.bisq.mobile.presentation.common.ui.utils.RememberPresenterLifecycleBackStackAware
+
+@Composable
+fun FaqScreen() {
+    val presenter = RememberPresenterLifecycleBackStackAware<FaqPresenter>()
```
rodvar, line 49: nit - for this static screen, RememberPresenterLifecycle is preferred

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/faqs/FaqScreen.kt
```diff
@@ -0,0 +1,255 @@
...
+}
+
+@Composable
+fun FaqScreenContent(
+    uiState: FaqUiState,
+    onAction: (FaqUiAction) -> Unit,
+    initialExpandedIndex: Int = -1,
+    topBar: @Composable () -> Unit = {},
+) {
+    var expandedIndex by remember { mutableIntStateOf(initialExpandedIndex) }
```
rodvar: nit - `rememberSaveable` would be preferred for this case?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/faqs/FaqPresenter.kt
```diff
@@ -0,0 +1,51 @@
...
+import network.bisq.mobile.presentation.common.ui.base.BasePresenter
+import network.bisq.mobile.presentation.common.ui.utils.BisqLinks
+import network.bisq.mobile.presentation.main.MainPresenter
+
+abstract class FaqPresenter(
+    mainPresenter: MainPresenter,
+    private val q2AnswerKey: String,
+    private val q3AnswerKey: String,
+) : BasePresenter(mainPresenter) {
+    private val _uiState = MutableStateFlow(FaqUiState(faqs = buildFaqList()))
```
rodvar, line 15: please fix: buildFaqList() only called at construction, presenter should also listen for language changes events and rebuild it into the uiState 🙏 

see `DashboardPresenter#launchLanguageChangeListenerJob` for an example

rodvar (changes requested):
Looking really good @salottodev !

Only one blocker to merge, and please review nits as well and apply if agreed 🙏

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/faqs/FaqPresenter.kt
salottodev: I think this change is not needed, this presenter will always be built with the right language. To change the language of the app you will have to go to Settings, if you want to navigate back to FAQs a new presenter will be built.

rodvar: sorry - you are right; I'm testing it now let me finish and likely will merge! 👍

rodvar (changes requested):
as discussed please explore increasing font size/spacing for the FAQs answers to improve UX

maybe SmallLight (14) → BaseLight/BaseRegular (16sp) ?

rodvar:
moved to next milestone

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/faqs/FaqScreen.kt
salottodev: @rodvar may I know why? I think we should start using RememberPresenterLifecycleBackStackAware, it is safer.

rodvar: > [@rodvar](https://github.com/rodvar) may I know why? I think we should start using RememberPresenterLifecycleBackStackAware, it is safer.

Of course, good question -  I just reread the doc I put in `LifecycleHelper.kt` @Composable header doc, it tells you what it was thought for but not the why (could be improved is what I mean)

`FaqPresenter` doesn't do any data fetching, is what's called a "static presentation". There's nothing whose state we need to preserve across back-stack navigation or config changes and that's why using the `RememberPresenterLifecycleBackStackAware` is not needed or overkill there.

That's the general answer, but specifically here, we are rebuilding i18n on construction. It's not a problem now but in the future if we were to provide a "use system language" setting and the language change, then using the state preservation would mean a bug. A little bit far-fetched but just thinking outloud here.

Anyways I would say this is a nit at least for now, feel free to change on a follow up or ping me to discuss!

rodvar (approved):
ACK

👏

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/settings/faqs/FaqScreen.kt
salottodev: @rodvar agreed this stays a nit and doesn't block — but let me make the broader case, because the FAQ line is the small end of something worth aligning on.
  
  **The reason I'd default to `BackStackAware` is concrete state loss.** With our `factory` Koin bindings, `RememberPresenterLifecycle` hands back a fresh presenter every time you return to a screen — so any
  state the presenter holds is lost on deep navigation (A → B → back to A), and on every config change. Choosing the helper per-screen based on whether it holds state *today* rots: a screen grows state in a
  few months, silently starts losing it on back navigation, and nobody remembers to switch helpers. That's exactly the regression that motivated `BackStackAware` in the first place. Under the hood this is
  just giving our presenters the same lifecycle parity as Android `ViewModel`s — retained while their entry is on the back stack, surviving config changes, cleared only on a real pop.
  
  **On "overkill" — I don't think it is.** `BackStackAware` is actually less code at the call site (one generic call instead of `koinInject()` + passing the presenter), and at runtime it's just a lightweight
  ViewModel holder around an already-cheap presenter. The cost is negligible, so it's not trading real overhead for safety. If anything, preserving state should be the default, and opting *out* to a fresh
  start is what should need justifying.

  **On the i18n concern — credit where due, it isn't actually stale today.** For FAQ you have to pop the screen to reach the language setting (it lives in Settings, FAQ's parent), so re-entry rebuilds the
  presenter in the new language. The staleness only appears if a presenter that bakes `.i18n()` into its constructor stays *retained on the back stack* while the language changes — a future "use system
  language" / OS-locale path, or changing language from a forward destination. It's latent, not present. And the fix is orthogonal to the lifecycle helper: our language switch already works via composable
  recomposition (`App` collects `languageCode` → tree recomposes → body `.i18n()` re-resolves; it's how `SettingsScreen` updates live). So the durable rule is **keys/structured data in presenters, `.i18n()` 
  in composables** — which makes localization correctness independent of the lifecycle choice entirely.

  **On retiring `RememberPresenterLifecycle` — gradually, not now.** I'd make `BackStackAware` the documented default and treat `RememberPresenterLifecycle` as an explicit, deprecated opt-out, then migrate
  existing screens over time. No big-bang change.

  **Concretely:**
  1. Make `BackStackAware` the documented default (lint-nudged), `RememberPresenterLifecycle` a deprecated opt-out.
  2. Rule: no `.i18n()` in presenters — keys in presenters, resolution in composables.
  3. Migrate presenters incrementally.

Does this direction work for you?

rodvar: Great points, mostly agree on the destination, let me put it in plain words:

  Fully on board:
  
  - The i18n rule is the real fix here 💪 : keys/structured data in presenters, .i18n() in composables. Makes localization correct independent of lifecycle. I agree on making it the documented convention 👍 
  - Gradual > big-bang - 👍 
  - And yeah, the doc @Composable header needs the why, not just the what — my bad, will improve it and I'll take the above as well 👍 

Things I think we could discuss for the future:

  1. The "presenters silently lose state" premise doesn't fully map to us --> we externalize durable/cross-nav state to coordinators (CreateOfferCoordinator/TakeOfferCoordinator), repos and services on purpose. What I mean is: presenters are thin projections where fresh-on-entry is a feature (re-subscribe/re-derive). The honest gap for BackStackAware is narrower: transient view state not worth a coordinator (scroll, expand/collapse, partial input) + config-change survival — which is actually the strongest motivator. 
  2. Flipping default carries risk, we need to review each case separately (this supports the "gradual > big-bang" approach we were talking about). Currently the majority of presenters uses the original approach including the Create/Take offer workflows. Removing that means onViewAttached only gets called once so there is a refactor to be done (where do we move that code kinda thing)
  3. i18n rule: this is a major effort, same as the above --> great as a forward convention, but let's apply it opportunistically, not as a mandated 50+ file work (same gradual > big-bang thinking)

So to summarise:

  1. Adopt the i18n rule now (docs + apply going forward).
  2. Keep fresh-start default for now, but fix the doc with the real decision rule so the choice is principled not incidental ( and mark it as deprecated?)
  3. ask crai to create an issue to audit the presenters using the old method and do the change for all the "Cheap" cases first? (would you like to take that?)
  4. FAQ stays RememberPresenterLifecycle (correct under both once i18n moves to the composable) 👍 
  
  what do you think? let's chat later today 😄

salottodev: Yeah, let's talk later today. Thanks

PR #1551, opened 2026-06-30
feat(bootstrap): redesign node splash with bootstrap progress, Tor failure and slow path handling

### LINKED ISSUE #434: Bootstrap: Mobile adaptation of new Bisq2 Desktop bootstrap UI
Image
Image

salottodev:
Resolves: https://github.com/bisq-network/bisq-mobile/issues/434

## Summary

* **New Features**
  * Added dedicated client and node splash screens with progress, status, and confirmation dialogs.
  * Node splash now shows step-by-step bootstrap flow, optional Tor recovery actions, and a slow-path banner.
  * Introduced deferred i18n rendering and new mobile startup localization strings.
* **Bug Fixes**
  * Updated splash navigation wiring so each app renders the correct splash destination; removed the outdated shared splash route.
  * Improved bootstrap failure handling to stop pending timeout/progress timers appropriately.
* **Tests**
  * Added UI and facade test coverage for splash/dialog behavior and bootstrap phase/progress timing.


apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashPresenter.kt
```diff
@@ -31,29 +36,221 @@ class NodeSplashPresenter(
...
+
+    private fun buildNodeUiState(
+        splashUiState: SplashUiState,
+        phase: BootstrapPhase,
+        torProgress: Int,
+        peerCount: Int,
+        dataReceived: Boolean,
+        appBootstrapFailed: Boolean,
+    ): NodeSplashUiState {
+        val torFailed = splashUiState.activeDialog == SplashActiveDialog.TorBootstrapFailed
```
rodvar: **Please review** — `torFailed` is derived from `activeDialog` here, which couples the node's whole Tor-failure UI (title, step status, inline actions) to `resolveActiveDialog`'s priority in the base presenter, where `isBootstrapFailed` outranks `torBootstrapFailed`. It works today, but reordering that `when` (or adding a higher-priority dialog) would silently break the node's Tor-failure path. I think the source of truth `applicationBootstrapFacade.torBootstrapFailed` is available --> prefer sourcing it directly (a nested combine, since the typed `combine` is already at its 6-arg max..?), or at minimum add a comment documenting the dependency on dialog priority 🙏

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashPresenter.kt
```diff
@@ -31,29 +36,221 @@ class NodeSplashPresenter(
...
-                numConnections,
-            ) { state, numConnections ->
-                if (numConnections >= 0) {
-                    "mobile.splash.bootstrapState.stateAndNumConnections".i18n(
-                        state,
-                        numConnections,
+            // 6-arg combine: use the project's typed combine (kotlinx only types up to 5);
+            // imported kotlinx.combine stays for the 3-arg slow-path collector below.
+            network.bisq.mobile.domain.utils
+                .combine(
```
rodvar: **Nit** we prefer full naming only in imports 🙏

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashPresenter.kt
```diff
@@ -31,29 +36,221 @@ class NodeSplashPresenter(
         settingsServiceFacade,
         versionProvider,
     ) {
-    private val _state = MutableStateFlow("")
-    override val state: StateFlow<String> = _state.asStateFlow()
-    val numConnections: StateFlow<Int> = networkServiceFacade.numConnections
+    private val nodeApplicationBootstrapFacade = applicationBootstrapFacade as NodeApplicationBootstrapFacade
```
rodvar: **Nit** — hard cast `as NodeApplicationBootstrapFacade`. The constructor could take `NodeApplicationBootstrapFacade` directly and pass it up to `super(...)` (which accepts the base type), removing the cast and its ClassCastException surface.

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashPresenter.kt
```diff
@@ -31,29 +36,221 @@ class NodeSplashPresenter(
...
+            ),
+        )
+
+    private fun torStatus(
+        phase: BootstrapPhase,
+        torFailed: Boolean,
+    ): NodeBootstrapStepStatus =
+        when {
+            torFailed -> NodeBootstrapStepStatus.FAILED
+            phase.ordinal >= BootstrapPhase.INITIALIZE_NETWORK.ordinal -> NodeBootstrapStepStatus.DONE
```
rodvar: **Nit** - ordinal-based comparison (`phase.ordinal >= INITIALIZE_NETWORK.ordinal`) is fragile to enum reordering. A `when` or an explicit `order` property would be reorder-safe. Minor, since the enum is intentionally ordered, flagging for durability.

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashScreen.kt
```diff
@@ -0,0 +1,557 @@
...
+                isDone ->
+                    BisqText.XSmallMedium(
+                        text = "✓",
+                        color = BisqTheme.colors.backgroundColor,
+                    )
+
+                isFailed ->
+                    BisqText.XSmallMedium(
+                        text = "✕",
+                        color = Color.White,
```
rodvar: **Nit** — `Color.White` here vs `BisqTheme.colors.backgroundColor` in the `isDone` branch above; use `BisqTheme.colors.white` for theme consistency.

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashUiState.kt
```diff
@@ -0,0 +1,31 @@
...
+    val showTorFailureActions: Boolean = false,
+)
+
+data class SlowPathUiState(
+    val isVisible: Boolean = false,
+    val elapsedSeconds: Long = 0L,
+)
+
+data class NodeBootstrapStep(
+    val icon: UiString,
```
rodvar: **Observation** — step icons are i18n keys resolving to glyphs, so the glyphs live in the translation files and a translator could alter/mangle them. Consider whether icons belong in i18n vs a constants object. Not blocking.

rodvar (changes requested):
code-based review only on first pass.

Overall: genuinely well-engineered +| the concurrency is carefully reasoned and commented, the navigation refactor is correct (well done 💪 )

Only one request for now, and the rest are all nits for your review. 

I did a quick test on node and the implementation of the design looks beautiful, but want to test it all more carefully. Will do soon and complete the review.

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashPresenter.kt
```diff
@@ -11,49 +11,292 @@ import network.bisq.mobile.data.service.settings.SettingsServiceFacade
...
+    private val numConnections: StateFlow<Int> = networkServiceFacade.numConnections
+    private val allDataReceived: StateFlow<Boolean> = networkServiceFacade.allDataReceived
 
     override fun onViewAttached() {
         super.onViewAttached()
 
+        // isBootstrapFailed / torBootstrapFailed are the source of truth for the two failure UIs.
+        // The typed combine below is already at its 6-arg max, so fold the two failure flags into a
+        // single input here rather than inferring Tor failure from SplashPresenter.resolveActiveDialog's
+        // dialog priority (which would silently break this path if that priority were reordered).
```
rodvar, line 59: excellent combine usage here

rodvar: **Nit** - I would refactor this into private methods to declaration of what onViewAttached setups is very clear. Non blocker

rodvar (approved):
ACK (node)

Also verified changes doesn't affect connect apps at all

Outstanding work, all concerns have been addressed + solution is robust and is visible in UX. 

Added one more nit. There is also a small UX hiccup when connection is unstable (up and down quickly) I don't think we should be concerned about that one (tor dialog gets shown twice quickly)


Image
Image
Image
Image
Image

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/startup/splash/NodeSplashPresenter.kt
```diff
@@ -11,49 +11,292 @@ import network.bisq.mobile.data.service.settings.SettingsServiceFacade
...
+            ),
+            NodeBootstrapStep(
+                icon = NodeBootstrapStepIcon.DATA,
+                label = UiString("mobile.bootstrap.node.step.data"),
+                detail = dataDetail(phase, dataReceived),
+                status = dataStatus(phase, dataReceived),
+            ),
+            NodeBootstrapStep(
+                icon = NodeBootstrapStepIcon.READY,
+                label = UiString("mobile.bootstrap.node.step.ready"),
```
salottodev, line 170: @coderabbitai when you review code next time and you see string keys calling .i18n() in presenters you should flag this. The way we should pass string to composable is through UiString class. Is this clear?

Issue #1514, opened 2026-06-18
Bisq Connect 0.6.2


rodvar:
- Smarter notifications: push notifications now open the relevant screen when you tap them.
  - Offers filter remembered: your selected market filter in the Offers tab now persists between sessions.
  - Steadier connections over Tor: fewer false "reconnecting" states thanks to Tor-aware timing.
  - **iOS** Clearer reconnection feedback: reconnect services functionality
  - **Android** Clearer reconnection feedback: Overlay shows every time the system is reconnecting
  - Plus other bugfixes, UX and performance improvements

rodvar:
moving as first task of next milestone to close this one

rodvar:
Issue renamed to .2 , version 0.6.1 wouldn't pass internal see issues above)

rodvar:
https://github.com/bisq-network/bisq-mobile/releases/tag/connect_0.6.2

PR #1564, opened 2026-07-02
bugfix/connect_first_market_fetch_after_cold_start_fails_freq_over_tor

### LINKED ISSUE #1563: Offerbook: markets show offers but empty when selected after cold start
**pre**: app is killed, paired with a tor trusted node

**steps**:

 1. open the app
 2. go to offers, observe markets with offers. Tap on any market with offers
 3. observe market open with spinner. Wait

**ER**: spinner finishes and offers are displayed

**OR**: spinner finishes, no offers displayed. Going back and forth to the same market retriggers the request and after 1-2 attempts it loads. Then it will always work.

**tech comment**: this can't happen in node because its reusing cached impl on bisq2 jars. We might want to consider same for connect (one solution) or a smart logic (if the app already knows market offers > 0, wait longer instead of timeouting or auto-retry ?).

rodvar:
- fixes #1563 
 - contribs to #1514 
 - fixed using TDD to avoid regressions int he future, increased timeout for loading market offers, improved logging and docs. Do not cancel the offers job early (~12 secs at the moment when roundtrip timeout is 30

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offerbook loading logic to be NUM_OFFERS-aware, keeping the spinner on for “unknown” or delayed OFFERS and clearing it only when counts indicate no offers.
  * Increased offer loading timeout and changed it to stop only the spinner while keeping live offers updates active for late snapshots.
  * Added REST prefetch as a best-effort path, with cache reconciliation and subscription data taking priority.

* **New Features**
  * Added a slow-loading hint snackbar for prolonged offer loading.

* **Tests**
  * Expanded coverage for OFFERS/NUM_OFFERS edge cases, REST prefetch and failure handling, and added slow-loading hint tests.

* **Chores**
  * Updated CI Gradle invocation to disable build cache.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,7 +160,11 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu')
         env:
           CI: true
-        run: ./gradlew clean test --no-configuration-cache -x :apps:nodeApp:test --info
+        # --no-build-cache forces test tasks to actually execute rather than being restored as
+        # build-cache hits. Kover's per-test coverage (.ic) binary is only produced when the test
+        # JVM runs with the Kover agent; a cache-hit test task produces no coverage, which then
+        # makes the separate koverVerify step see near-zero coverage intermittently (flaky failures).
+        run: ./gradlew clean test --no-configuration-cache --no-build-cache -x :apps:nodeApp:test --info
 
       - name: Run Kover Verification (Ubuntu only, excluding nodeApp)
         if: startsWith(matrix.os, 'ubuntu')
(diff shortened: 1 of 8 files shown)


rodvar:
issue and UX is greatly improved, but I can still repro it (only happens on the first load after cold start now) - I'm working to tackle that last case and then merge

rodvar:
> issue and UX is greatly improved, but I can still repro it (only happens on the first load after cold start now) - I'm working to tackle that last case and then merge

this is part of a bigger issue... I think what we need to do to fully solve that edge case is:

  1. Don't block the chain on each subscription — fire subscribe requests without awaiting each response serially (apply concurrently, or at least don't let one 30s timeout gate the rest).
  2. Prioritize OFFERS in subscriptionEntriesInApplyOrder (it's currently after the banner set and after TRADES/CLOSED_TRADES).
  3. Fast-path via the dead REST endpoint: OfferbookApiGateway.getOffers(code) (offerbook/markets/$code/offers) already exists but is unused.
 
@coderabbitai could you please create an issue to address the above soon

In the meantime I'll explore this: On market select we could fire that REST call to fetch just the selected market's offers immediately — REST req/resp doesn't sit behind applySubscriptions, so it'd return in seconds even while the OFFERS subscription is stuck. If that works with no big effort I'll add it in will bring best UX

rodvar:
rebased and improved tests

rodvar:
merged to include in connect version 0.6.2 (0.6.1 will never see the light of day)

PR #1573, opened 2026-07-03
refactor(offerbook): bind market presenter as factory to migrate the screen to UiState/UiAction

### LINKED ISSUE #1304: Refactor OfferbookMarketPresenter/ClientOfferbookPresenter/NodeOfferbookPresenter to make stateless & enable factory instantiation
## Background

`OfferbookMarketPresenter`, `ClientOfferbookPresenter`, and `NodeOfferbookPresenter` are kept as
singletons because they intentionally preserve filter/sort state across tab switches. Before these
can be converted to factory instances (on-demand instantiation), the state management strategy
needs to be refactored so that filter/sort state is maintained independently of the presenter
lifecycle.

## Task

Refactor the state-preservation mechanism for offerbook filter/sort state so that it is decoupled
from the presenter singleton lifecycle, then convert these presenters from singleton to factory
instantiation to align with the pattern established in `#1303`.

## References

- Tracked as part of the singleton-to-factory migration: `#1303`

salottodev:
Partially resolves: https://github.com/bisq-network/bisq-mobile/issues/1304

## Summary by CodeRabbit

* **New Features**
  * The offerbook market tab now uses a more unified interaction flow for search, filtering, sorting, and market selection.
  * Market selection now shows an error message if navigation can’t be completed.

* **Bug Fixes**
  * Search, filter, and sort choices now stay consistent while moving between screens and returning to the tab.
  * User preferences for market view settings are preserved more reliably across updates.


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/di/PresentationModule.kt
```diff
@@ -145,7 +145,7 @@ val presentationModule =
         factory { ComputeOfferbookMarketListUseCase(get()) }
 
         // Offerbook
-        single<OfferbookMarketPresenter> { OfferbookMarketPresenter(get(), get(), get(), get(), get(), get(), get()) }
+        factory<OfferbookMarketPresenter> { OfferbookMarketPresenter(get(), get(), get(), get(), get(), get(), get()) }
```
rodvar, line 148: 💪

shared/presentation/src/androidUnitTest/kotlin/network/bisq/mobile/presentation/tabs/offers/OfferbookMarketPresenterLifecycleTest.kt
```diff
@@ -1,306 +0,0 @@
...
-            // Sort should still be changeable after tab switch
-            presenter.setSortBy(MarketSortBy.NameZA)
-            advanceUntilIdle()
-
-            assertEquals(MarketSortBy.NameZA, presenter.sortBy.value)
-            assertEquals(MarketSortBy.NameZA, settingsRepository.fetch().marketSortBy)
-        }
-
-    @Test
-    fun `filter reflects DataStore value on attach`() =
```
rodvar, line 174: the remove/add of tests is very well done in general and makes sense from the approach change perspective.

But this test (we could call it a `restore-from-persistence read-path`) has lost is coverage in the new set of tests. Could we please consider adding it? not a blocker since this is a first of a series of PRs to kill the related issue, please consider adding next 🙏

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/tabs/offers/OfferbookMarketPresenter.kt
```diff
@@ -34,105 +33,67 @@ class OfferbookMarketPresenter(
...
-    }
-
-    private fun observeGlobalMarketPrices() {
-        presenterScope.launch {
-            marketPriceServiceFacade.globalPriceUpdate.collect { timestamp ->
-                log.d { "Offerbook received global price update at timestamp: $timestamp" }
-                val previousValue = _marketPriceUpdated.value
-                _marketPriceUpdated.value = !_marketPriceUpdated.value
-                log.d { "Offerbook triggered market filtering update: $previousValue -> ${_marketPriceUpdated.value}" }
+                showSnackbar("mobile.error.generic".i18n(), type = SnackbarType.ERROR)
```
rodvar, line 96: nit - great we i18n but the generic errors loose the context "failed to select market.." was the previous error

I don't think is a great deal since this would happen immediately after main action of selecting market, and also without selecting market nothing elase con be done in this workflow there is a nit

rodvar (approved):
ACK

Issue #1424, opened 2026-05-12
Bootstrap: control port unaccesible not being informed, user stuck at tor starting


rodvar:
When this happens user sees "Tor Starting 10%" stucked forever until the stall detection striggers but solutions offered wouldn't fix this case (app restart). If the user restarts the device (fully frees the port) everything works normally.

Apply one solution or propose a new one:

 - auto swap port?
  - Bubble Tor bootstrap failure (or stall > N seconds at the same %) up to the splash UI as an error state with "Retry" / "Show logs" actions, instead of just logging and continuing
  - The KmpTorService warning should arguably surface as an ErrorState flow the splash presenter consumes 

```
2026-05-12 12:52:57.093  5712-5712  KmpTorService           network.bisq.mobile.node.debug       W  Control port 33979 not yet accessible, but continuing anyway
```

rodvar:
this one needs retesting after the work we merged today @salottodev , it might not be happening anymore

PR #1579, opened 2026-07-06
chore/removal_of_dead_code_post_refactors

rodvar:
- can't repro #1424 anymore we can close it now
 - remove dead code whilst verifying tor control port block issue not happening after latest node bootstrap redesign and refactor

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified node startup and splash behavior so failures are shown consistently through the main application state.
  * Removed Tor-specific failure handling from the splash flow, preventing extra failure actions and confusing error states.
  * Updated the splash screen to show a more stable progress experience during startup.


rodvar:
We had a concern regarding inlining of tor slow warnings during the bootstrapping process.

I verified this PR removes dead-code-only. The node never starts `KmpTorService` (bisq2 core owns Tor — startTor() is only called Connect-side, per NodeDomainModule and Bisq2SocksPortProvider), so **torBootstrapFailed** is always **false** and **torBootstrapProgress** always **0**.

That makes all the removed Tor-failure UI (TorFailureActions, torFailed titles/steps, the TorBootstrapFailed dialog suppression, observeTorState()) unreachable; and the restart/clear-Tor buttons even called startTor() on a service the node never uses, so they'd have been broken if shown. Real Tor/network failures still surface via bisq2 core State.FAILED → setBootstrapFailed(true). Importantly, the slow-connection "still connecting, this is normal" banner (SlowPathBanner) is fully preserved (I've just retested this ✅ ) 

Finally, BaseModel has zero references repo-wide.

PR #1585, opened 2026-07-08
Fixes splash getting stuck at "Starting Tor… 0%" when kmp-tor dies early with CtrlConnection Stream Ended

### LINKED ISSUE #1572: iOS: Tor stuck at 0% with no recovery after CtrlConnection Stream Ended
On iOS (seen intermittently on the simulator at app start), the tor daemon dies early during bootstrap with `CtrlConnection Stream Ended`. The app then sits at **Tor 0% forever with no dialog** and no way to recover.

## Symptom
- Splash stuck at "Starting Tor… 0%".
- No "Restart services" / Tor-failed dialog appears.
- Logs:
  ```
  Bootstrap: Tor error within grace period (Ns): CtrlConnection Stream Ended — suppressing failure dialog
  🔴 (KmpTorService) Starting kmp-tor daemon failed: CtrlConnection Stream Ended
  io.matthewnelson.kmp.file.InterruptedException: CtrlConnection Stream Ended
      at ...RealTorCtrl.Processor...awaitReplies... (runBlocking on MultiWorkerDispatcher)
  ```

## My analysis
Two compounding issues create a deadlock:

1. **Terminal failure treated as transient.** `CtrlConnection Stream Ended` means the control connection died and the daemon is stopped (`TorState.Stopped(error)`) — terminal. But it fails within the 60s window, so `ApplicationBootstrapFacade.observeTorState()` takes the grace-period branch, logs "suppressing failure dialog", repaints 0%, and **does not** set `torBootstrapFailed`. The comment says "let Tor retry" but nothing re-invokes `startTor()`.
   - `ApplicationBootstrapFacade.kt:196` (grace-period suppress branch)
2. **Recovery path is unreachable.** The restart dialog (`SplashDialogs.kt:` → `OnRestartTor`/`OnPurgeRestartTor`, the only code path that calls `startTor()` again) is gated on `torBootstrapFailed`, which was never set. Meanwhile `NetworkServiceFacade.activate()` suspends forever on `state.filter { Started }.first()` despite its "we restart tor till it starts" comment (there is no retry).

The likely trigger for the daemon death itself is cancellation-shaped (`InterruptedException` + interrupted `runBlocking`): a lifecycle scope cancellation mid-bootstrap, a stale tor data-dir/lock from a prior simulator relaunch, or simulator loopback flakiness.

## Impact
- App unusable (permanent 0%) with no user-facing recovery when this fires. (only options is hard kill and restart)
- Likely existing workaround: background + re-foreground the app (triggers `ensureTorRunning()`), to confirm.

## Suggested fix
- Classify `Stopped(error)` — terminal signals (control-connection ended / daemon stopped / non-cancellation interrupts) should bypass the grace period and arm `torBootstrapFailed` (show the restart dialog).
- Make "let Tor retry" real: the suppress branch should actually re-invoke `startTor()` (bounded retries) and only escalate to the dialog after N failures / after 60s.
- Investigate whether an app-lifecycle `deactivate()`/reactivation cancels the bootstrap scope during startup on the simulator.

nostrbuddha:
Fixes https://github.com/bisq-network/bisq-mobile/issues/1572

After multiple attempts, I couldn't get the "CtrlConnection Stream Ended" error. So I added the error simulation code in kmpTorService, observed the issues and fixed them.

 - Classifies terminal Tor errors (CtrlConnection Stream Ended, daemon stopped, etc.) and throws torBootstrapFailed immediately, bypassing the grace period. With `TorBootstrapErrorClassification`, we can now define which errors are terminal and which are transient that can be retried.
 - Transient errors within 60s trigger bounded startTor() retries (up to 3) before escalating to the failure dialog.
 - Unblocks NetworkServiceFacade.activate() — waits for either Tor started or bootstrap failed, then throws TorBootstrapNotReadyException (swallowed as expected failure, not unrecoverable).

androidNode specific:
 - Wires splash recovery - "Restart Tor" / "Purge & restart" run a full in-process lifecycle restart via restartTorBootstrap(), required after activation aborts (especially on Node, where bisq2 may never have left INITIALIZE_APP) - Without this, after retry Tor reached 100% connectivity and connection to p2p network is not initialized.

To quickly simulate the error and test the fixes, you can undo the cleanup logic in 5cee534c9f0e45a8998096d08c917aca295d4f69

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Tor bootstrap restart flow with automatic recovery, including an option to retry after purging Tor data.
  * Exposed Tor restart control via the main presenter and updated the splash recovery flow to use it.
* **Bug Fixes**
  * Improved handling of Tor bootstrap readiness failures as expected recovery events, with better terminal vs non-terminal detection.
  * Enhanced lifecycle activation/shutdown behavior during startup to avoid interrupting early recovery attempts.

### DIFF
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
@@ -6,7 +6,7 @@ import kotlinx.coroutines.CancellationException
  * Classifies Tor bootstrap [Throwable]s for grace-period handling in [ApplicationBootstrapFacade].
  *
  * Terminal failures must surface [ApplicationBootstrapFacade.torBootstrapFailed] immediately.
- * Transient circuit-establishment errors may be retried during the grace window (P1).
+ * Transient circuit-establishment errors are suppressed until the grace window elapses.
  */
 object TorBootstrapErrorClassification {
     fun isTerminal(error: Throwable): Boolean {
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
@@ -1,0 +1,41 @@
+package network.bisq.mobile.data.service.bootstrap
+
+import kotlinx.coroutines.CancellationException
+
+/**
+ * Classifies Tor bootstrap [Throwable]s for grace-period handling in [ApplicationBootstrapFacade].
+ *
+ * Terminal failures must surface [ApplicationBootstrapFacade.torBootstrapFailed] immediately.
+ * Transient circuit-establishment errors may be retried during the grace window (P1).
+ */
+object TorBootstrapErrorClassification {
+    fun isTerminal(error: Throwable): Boolean {
+        if (error is CancellationException) {
+            return false
+        }
+        return collectMessages(error).any(::isTerminalMessage)
+    }
+
+    internal fun isTerminalMessage(message: String): Boolean {
+        val normalized = message.lowercase()
+        return TERMINAL_MESSAGE_MARKERS.any { marker -> normalized.contains(marker) }
+    }
+
+    private fun collectMessages(error: Throwable): List<String> {
+        val messages = mutableListOf<String>()
+        var current: Throwable? = error
+        while (current != null) {
+            current.message?.let { messages.add(it) }
+            current = current.cause
+        }
+        return messages
+    }
+
+    private val TERMINAL_MESSAGE_MARKERS =
+        listOf(
+            "ctrlconnection stream ended",
+            "daemon stopped",
+            "control connection",
+            "tor process exited",
+        )
+}
--- a/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
+++ b/shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
@@ -7,6 +7,10 @@ import kotlinx.coroutines.CancellationException
  *
  * Terminal failures must surface [ApplicationBootstrapFacade.torBootstrapFailed] immediately.
  * Transient circuit-establishment errors are suppressed until the grace window elapses.
+ *
+ * String matching notes (kmp-tor has no structured failure kinds today):
+ * - Markers validated against kmp-tor **2.6.0** and kmp-tor-resource **409.5.0** (`gradle/libs.versions.toml`).
+ * - Re-check [TERMINAL_MESSAGE_MARKERS] after any kmp-tor upgrade — wording can change between releases.
  */
 object TorBootstrapErrorClassification {
     fun isTerminal(error: Throwable): Boolean {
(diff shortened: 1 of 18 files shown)


rodvar:
@coderabbitai full review

rodvar:
@nostrbuddha can you check the above CRAI review please before my review? 🙏

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/network/NetworkServiceFacade.kt
```diff
@@ -62,4 +63,41 @@ abstract class NetworkServiceFacade(
...
+                throw TorBootstrapNotReadyException()
+            }
+        }
+    }
+
+    private sealed interface TorActivationOutcome {
+        data object Started : TorActivationOutcome
+
+        data object BootstrapFailed : TorActivationOutcome
+    }
```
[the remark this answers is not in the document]
nostrbuddha: This is the intended behavior. Setting `torBootstrapFailed` on timeout would be wrong.

90s timeout = Tor is still in progress but slower than the stage timer expected. The timeout dialog is shown, with the option to Continue waiting. If the user taps Continue, extendTimeout() resets the timer and everything keeps waiting. 90s wait doesn't mean tor failed.

awaitTorStartedOrBootstrapFailed() waits only for Tor started, or failed. In case of 90s timeout, it just waits.

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/network/NetworkServiceFacade.kt
```diff
@@ -31,6 +38,7 @@ abstract class NetworkServiceFacade(
 
         if (isTorEnabled()) {
             startTorWithRetries()
+            awaitTorStartedOrBootstrapFailed()
```
rodvar, line 48: please fix: add a sensible timeout to this await ( can reuse existing timeout vals for Tor)

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import kotlinx.coroutines.CancellationException
+
+/**
+ * Classifies Tor bootstrap [Throwable]s for grace-period handling in [ApplicationBootstrapFacade].
+ *
+ * Terminal failures must surface [ApplicationBootstrapFacade.torBootstrapFailed] immediately.
+ * Transient circuit-establishment errors are suppressed until the grace window elapses.
+ */
+object TorBootstrapErrorClassification {
```
rodvar, line 15: 👏

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/bootstrap/TorBootstrapErrorClassification.kt
```diff
@@ -0,0 +1,42 @@
...
+    fun isTerminal(error: Throwable): Boolean {
+        if (error is CancellationException) {
+            return false
+        }
+        return collectMessages(error).any(::isTerminalMessage)
+    }
+
+    internal fun isTerminalMessage(message: String): Boolean {
+        val normalized = message.lowercase()
+        return TERMINAL_MESSAGE_MARKERS.any { marker -> normalized.contains(marker) }
```
rodvar, line 25: I understand its a challenge to address how kmp-tor handles errors and hence why we are relying on these wording to compare. Good you added the lowercase standarization

However, this is a flaky approach given how the lib changes copy across versions. I worry next kmp-tor upgrade would break this.

Can we add a comment here at least specifying which kmp-tor version this was tested and is compatible with? If you feel like itwould be good to ship an issue for kmp-tor devs to expose this somehow?

rodvar (changes requested):
is looking really good buddha, please check my comments and then after addressed I'll git it a final test and merge 💪

rodvar:
rebased to main

rodvar (approved):
ACK

all concerns addressed 👍

Issue #1524, opened 2026-06-20
Settings: Network Info with already accesible data


rodvar:
## Menu

<img width="611" height="690" alt="Image" src="https://github.com/user-attachments/assets/054f5810-7e8d-4c55-ad2d-7dfafbf5c809" />

## Node

<img width="918" height="675" alt="Image" src="https://github.com/user-attachments/assets/15baa616-84c6-4e05-864b-eccf239af5f7" />

<img width="921" height="685" alt="Image" src="https://github.com/user-attachments/assets/8a1c8998-8b43-4cc5-95b9-1d850e80ed84" />

## Connect

<img width="935" height="683" alt="Image" src="https://github.com/user-attachments/assets/2ac98d8b-1f3e-4643-87d7-7f836b42fa34" />

<img width="941" height="684" alt="Image" src="https://github.com/user-attachments/assets/b83196c3-f237-439d-aad7-873cee88d8ab" />

salottodev:
I can take this one

PR #1586, opened 2026-07-09
design/clarifications_on_network_new_feature_docs

rodvar:
- contrib to #1524 designs
 - clarifications on where to take the data from in both node and connect implementations, also fixed some "design bugs" referencing data that should be stale (node).


rodvar:
@salottodev please let me know if we need further clarifications whilst you develop the feature 🙏

PR #1601, opened 2026-07-13
Guard Transifex workflow_run uploads

hiciefte:
## Summary

- Limit the `workflow_run` Transifex trigger to `BisqApps` runs on `main`.
- Add job-level guards so Transifex uploads only follow successful `push` builds from the base repository's default branch.
- Keep the existing `pull_request` Transifex configuration check for i18n and `.tx/config` changes.
- Restrict workflow token permissions to read-only contents and avoid persisted checkout credentials.

## Why

Bisq Mobile currently still uses `actions/checkout@v6`, so it has not hit the same checkout guard that recently broke Bisq2 after the `checkout@v7` bump. The workflow has the same latent shape, though: a trusted `workflow_run` path checks out `github.event.workflow_run.head_sha` before proving that the completed `BisqApps` run came from a trusted `main` push.

If checkout is later upgraded to v7, fork PR builds can fail with the same protected checkout error seen in Bisq2. More importantly, the privileged Transifex upload path should not inspect or execute fork PR code with repository secrets available.

This mirrors the Bisq2 fix by skipping the privileged upload path for PR-triggered `workflow_run` events and allowing uploads only after trusted main push builds.

## Verification

- Parsed `.github/workflows/sync_transifex.yml` as YAML locally.
- Ran `git diff --check`.
- Reviewed recent Transifex and `BisqApps` runs: current Mobile Transifex runs are not failing from this yet, but the workflow shape matches the Bisq2 latent issue.

## Summary by CodeRabbit

* **Chores**
  * Improved translation synchronization workflow safeguards by restricting runs to successful updates from the default branch and repository.
  * Limited pull request-triggered synchronization to pull requests targeting `main`.
  * Enhanced checkout security by preventing credentials from being persisted in workflow environments.

### DIFF
--- a/.github/workflows/sync_transifex.yml
+++ b/.github/workflows/sync_transifex.yml
@@ -10,11 +10,24 @@ on:
   workflow_run:
     workflows: [ BisqApps ]
     types: [ completed ]
+    branches:
+      - main
+
+permissions:
+  contents: read
 
 jobs:
   verify_config:
     name: Verify Transifex configuration
-    if: github.event_name == 'pull_request' || (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')
+    if: >
+      github.event_name == 'pull_request' ||
+      (
+        github.event_name == 'workflow_run' &&
+        github.event.workflow_run.conclusion == 'success' &&
+        github.event.workflow_run.event == 'push' &&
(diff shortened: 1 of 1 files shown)


rodvar (approved):
utACK

PR #1636, opened 2026-07-26
Downsample CatHash bitmap layers during decode

### LINKED ISSUE #1578: GP recommendation on cathash service
Image

nostrbuddha:
Fixes https://github.com/bisq-network/bisq-mobile/issues/1578

- Subsample CatHash PNG layers during decode via `BitmapFactory.Options.inSampleSize` (300→150 at 120px target)
- Two-pass decode follows [Android’s load-bitmap guide](https://developer.android.com/topic/performance/graphics/load-bitmap): bounds pass with inJustDecodeBounds, then calculateInSampleSize() (power-of-two), then decode with inSampleSize set
- Pass `Options` on every CatHash-related `BitmapFactory.decode*` — compose, cache read, and `PlatformImage.deserialize` — to satisfy GP’s static check

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved image loading by decoding assets closer to their requested display size.
  * Reduced unnecessary memory usage when displaying smaller images.
  * Enhanced image composition to maintain correct dimensions and visible content across different sizes.

### DIFF
--- a/shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/utils/PlatformDomainAbstractions.android.kt
+++ b/shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/utils/PlatformDomainAbstractions.android.kt
@@ -174,9 +174,8 @@ actual class PlatformImage(
 ) {
     actual companion object {
         actual fun deserialize(data: ByteArray): PlatformImage {
-            val decodeOptions = BitmapFactory.Options()
             val bitmap =
-                BitmapFactory.decodeByteArray(data, 0, data.size, decodeOptions)
+                BitmapFactory.decodeByteArray(data, 0, data.size)
                     ?: throw IllegalArgumentException("Failed to decode image data")
             return PlatformImage(bitmap.asImageBitmap())
         }
--- a/shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/utils/PlatformDomainAbstractions.android.kt
+++ b/shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/utils/PlatformDomainAbstractions.android.kt
@@ -174,8 +174,9 @@ actual class PlatformImage(
 ) {
     actual companion object {
         actual fun deserialize(data: ByteArray): PlatformImage {
+            val decodeOptions = BitmapFactory.Options()
             val bitmap =
-                BitmapFactory.decodeByteArray(data, 0, data.size)
+                BitmapFactory.decodeByteArray(data, 0, data.size, decodeOptions)
                     ?: throw IllegalArgumentException("Failed to decode image data")
             return PlatformImage(bitmap.asImageBitmap())
         }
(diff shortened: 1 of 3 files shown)


shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/utils/PlatformDomainAbstractions.android.kt
```diff
@@ -174,8 +174,9 @@ actual class PlatformImage(
 ) {
     actual companion object {
         actual fun deserialize(data: ByteArray): PlatformImage {
+            val decodeOptions = BitmapFactory.Options()
```
rodvar: nit: do we need this? I think is a no-op..?

shared/presentation/src/androidMain/kotlin/network/bisq/mobile/presentation/common/ui/utils/AndroidImageUtil.kt
```diff
@@ -118,7 +164,10 @@ object AndroidImageUtil : Logging {
         return outputStream.toByteArray()
     }
 
-    fun decodePngToImageBitmap(pngByteArray: ByteArray): Bitmap = BitmapFactory.decodeByteArray(pngByteArray, 0, pngByteArray.size)
+    fun decodePngToImageBitmap(pngByteArray: ByteArray): Bitmap {
+        val decodeOptions = BitmapFactory.Options()
+        return BitmapFactory.decodeByteArray(pngByteArray, 0, pngByteArray.size, decodeOptions)
```
rodvar: nit - this fun returns non nullable Bitmap but the decodeByteArray can actually return null if there is corrupted data.

but I think this now has 0 callers? can you please check, if its dead code lets take the opportunity to remove?

rodvar (changes requested):
code looks great, follows Android stds.

Just a couple of nits, have a look and I'll give it a full test tomorrow

shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/utils/PlatformDomainAbstractions.android.kt
nostrbuddha: Yes. It's a no-op. Removing it!

shared/presentation/src/androidMain/kotlin/network/bisq/mobile/presentation/common/ui/utils/AndroidImageUtil.kt
nostrbuddha: Correct. Not just this. The following functions in that file is not used anymore:
 - readPngByteArray
 - saveByteArrayAsPng
 - bitmapToPngByteArray
 - saveBitmapAsPng

Removing them all.

rodvar (approved):
ACK

cool 👍 

I've done some logging fixes regarding job cancellation handling whilst testing this (its very noisy when there are lots of offers in a market). But I'll make a separate PR for it as its not that small to add here.

PR #1638, opened 2026-07-28
refactor: non-null AnalyticsService DI, consolidated test fakes

### LINKED ISSUE #1488: Refactor: replace nullable/lazy AnalyticsService workaround with proper DI and shared test utilities
## Context

This issue tracks follow-up improvements identified during PR #1486 (opt-in analytics wrap-up and dashboard promo).

Requested by @rodvar.

## Background

In PR #1486, `BasePresenter.analyticsService` was changed from an eager, non-null Koin injection to a `by lazy { runCatching { getKoin().getOrNull<AnalyticsService>() }.getOrNull() }` pattern. This was necessary because some UI-test harnesses construct presenters before `startKoin {}` is called, and `getKoin()` would throw `KoinApplication has not been started`. The `by lazy + runCatching` approach defers resolution and swallows that crash, leaving the service `null` in tests that don't bind it.

Additionally, the visibility of `analyticsScreenEvent()` was changed from `protected` to `internal` to allow screen-coverage contract tests in the same module to access it directly.

As a side effect, every `FakeSettingsRepository` in each test file had to be manually extended with `setAnalyticsBaselineSent` stubs.

## What to improve

1. **Replace the nullable/lazy `AnalyticsService` pattern with proper DI.**  The long-term solution is to ensure tests that exercise presenter logic either: (a) set up Koin with a no-op/mock `AnalyticsService` binding before constructing presenters, or (b) receive the service via constructor injection rather than Koin resolution inside the class. This removes the need for nullable access and `runCatching` swallowing in production code.

2. **Add a shared `NoOpAnalyticsService` (or `FakeAnalyticsService`) to `shared:test-utils`.**  Currently tests that don't care about analytics tracking get a silent null. A shared no-op implementation bound in a common test Koin module would make the DI contract explicit and avoid the nullable production path.

3. **Consolidate `FakeSettingsRepository` stubs.**  Multiple test files (e.g. `CreateOfferPricePresenterTest`, `CreateOfferReviewPresenterTest`, `TakeOfferCoordinatorTest`, `NodeOffersServiceFacadeIntegrationTest`) each contain their own copy of `FakeSettingsRepository` that needs updating every time `SettingsRepository` gains a new method. A single canonical `FakeSettingsRepository` in `shared:test-utils` (already partially done via `SettingsRepositoryMock`) should be the single source of truth used everywhere.

4. **Revert `analyticsScreenEvent()` visibility to `protected`** once the test-utils approach is in place and the contract test can use a proper DI-bound mock rather than internal reflection.

## References

- PR #1486
- Issue #525 (original opt-in analytics requirement)
- `shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/base/BasePresenter.kt` — see the `analyticsService` KDoc for full rationale

salottodev:
resolves: https://github.com/bisq-network/bisq-mobile/issues/1488

## Summary by CodeRabbit

* **New Features**
  * Screen-opened analytics events are now consistently recorded when views attach.
  * Analytics tracking is available across presenter flows, including settings, language, and trade phases.

* **Bug Fixes**
  * Improved reliability of analytics event delivery by ensuring required analytics services are available.

* **Documentation**
  * Updated testing guidance and shared test utility documentation for analytics and reusable test fixtures.

### DIFF
--- a/apps/clientApp/src/androidUnitTest/kotlin/network/bisq/mobile/client/common/di/TestModule.kt
+++ b/apps/clientApp/src/androidUnitTest/kotlin/network/bisq/mobile/client/common/di/TestModule.kt
@@ -11,6 +11,8 @@ import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import network.bisq.mobile.data.replicated.user.profile.UserProfileVO
 import network.bisq.mobile.data.service.network.ConnectivityService
 import network.bisq.mobile.data.utils.PlatformImage
+import network.bisq.mobile.domain.analytics.AnalyticsService
+import network.bisq.mobile.domain.analytics.NoOpAnalyticsService
 import network.bisq.mobile.domain.utils.CoroutineExceptionHandlerSetup
 import network.bisq.mobile.domain.utils.CoroutineJobsManager
 import network.bisq.mobile.domain.utils.DefaultCoroutineJobsManager
@@ -38,6 +40,10 @@ val clientTestModule =
 
         // Provide a test dispatcher-based GlobalUiManager
         single { GlobalUiManager(UnconfinedTestDispatcher()) }
+
+        // BasePresenter resolves AnalyticsService as a non-null dependency. Tests that assert on
+        // analytics load their own mock after this module, which overrides the no-op.
+        single<AnalyticsService> { NoOpAnalyticsService }
 
         // Provide a default NavigationManager stub
         single<NavigationManager> {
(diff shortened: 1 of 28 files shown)


apps/clientApp/src/androidUnitTest/kotlin/network/bisq/mobile/client/common/di/TestModule.kt
```diff
@@ -39,6 +41,10 @@ val clientTestModule =
         // Provide a test dispatcher-based GlobalUiManager
         single { GlobalUiManager(UnconfinedTestDispatcher()) }
 
+        // BasePresenter resolves AnalyticsService as a non-null dependency. Tests that assert on
+        // analytics load their own mock after this module, which overrides the no-op.
+        single<AnalyticsService> { NoOpAnalyticsService }
```
rodvar, line 46: 👍

rodvar (approved):
utACK

solid refactor, thanks for updating md docs files as well

PR #1641, opened 2026-07-28
chore/bisqeasy_trade_model_inmutability

### LINKED ISSUE #1560: Make BisqEasyTradeModel immutable (remove public MutableStateFlow usage)
## Summary

Follow-up to #518 ("Apply immutability patterns"), which is being closed as the original scope was too broad to track as a single, actionable issue.

This issue narrows the goal to the specific, concrete piece of work that was repeatedly identified in the discussion on #518 as the highest-value target: making `BisqEasyTradeModel` immutable by removing its publicly exposed `MutableStateFlow` properties.

## Current state (verified 2026-07-02)

`shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/replicated/trade/bisq_easy/BisqEasyTradeModel.kt` still exposes the following as public `MutableStateFlow` properties, allowing any consumer to mutate trade state directly instead of going through a controlled update path:

- `tradeState`
- `interruptTradeInitiator`
- `paymentAccountData`
- `bitcoinPaymentData`
- `paymentProof`
- `errorMessage`
- `errorStackTrace`
- `peersErrorMessage`
- `peersErrorStackTrace`
- `tradeCompletedDate`

## Required changes

- Convert the public `MutableStateFlow` properties on `BisqEasyTradeModel` to read-only `StateFlow` (e.g. via a private backing `MutableStateFlow` with a public `StateFlow` getter, or an equivalent immutable wrapper).
- Update all call sites that currently mutate these flows directly (e.g. `.value = ...`, `.update { }`, `.emit(...)`) to go through model-owned update methods instead.
- Audit usages across the codebase (repositories, presenters/view models, use cases) that read/write these fields to ensure they compile against the new read-only API.
- Once `BisqEasyTradeModel` is fully immutable from the outside, evaluate spreading the same pattern to similar classes as originally suggested in #518.

## Rationale

Discussion on #518 converged on this being the concrete, bounded starting point for the broader immutability effort, since the original issue title/scope was too generic to act on directly. Narrowing to this class keeps the change reviewable and testable, in line with the team's preference for small, targeted PRs before/around release milestones.

## Acceptance criteria

- [ ] `BisqEasyTradeModel` no longer exposes mutable `MutableStateFlow` properties publicly.
- [ ] All internal call sites are updated to use the new read-only API + explicit update methods.
- [ ] No behavioral regressions in trade state handling (manual/automated verification as appropriate).

## References

- Original issue: https://github.com/bisq-network/bisq-mobile/issues/518
- Discussion comment: https://github.com/bisq-network/bisq-mobile/issues/518#issuecomment-4857699797
- Requested by: @rodvar

rodvar:
- resolves #1560 
 - made BisqEasyTradeModel immutable
 - applied necessary changes in usages
 - added AI generated tests to ensure behaviour is kept

## Summary by CodeRabbit

* **Improvements**
  * Trade information updates are now handled more consistently, helping keep displayed status, payment details, errors, and completion information synchronized.
  * Observable trade data is protected from unintended external modification while remaining responsive to updates.
  * Added support for connecting observable values to update callbacks, including optional value mapping.

* **Tests**
  * Expanded coverage for observable updates, trade model initialization, read-only state exposure, and setter-driven changes.

### DIFF
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
@@ -346,15 +346,16 @@ class ClientTradesServiceFacade(
         data: TradePropertiesDto,
     ) {
         log.i { "Apply mutable data to trade with ID $tradeId - new state: ${data.tradeState}" }
+        val tradeModel = trade.bisqEasyTradeModel
         data.tradeState?.let {
-            log.i { "Updating trade $tradeId state from ${trade.bisqEasyTradeModel.tradeState.value} to $it" }
-            trade.bisqEasyTradeModel.tradeState.value = it
+            log.i { "Updating trade $tradeId state from ${tradeModel.tradeState.value} to $it" }
+            tradeModel.setTradeState(it)
         }
-        data.tradeCompletedDate?.let { trade.bisqEasyTradeModel.tradeCompletedDate.value = it }
-        data.paymentAccountData?.let { trade.bisqEasyTradeModel.paymentAccountData.value = it }
-        data.bitcoinPaymentData?.let { trade.bisqEasyTradeModel.bitcoinPaymentData.value = it }
-        data.paymentProof?.let { trade.bisqEasyTradeModel.paymentProof.value = it }
-        data.interruptTradeInitiator?.let { trade.bisqEasyTradeModel.interruptTradeInitiator.value = it }
+        data.tradeCompletedDate?.let { tradeModel.setTradeCompletedDate(it) }
+        data.paymentAccountData?.let { tradeModel.setPaymentAccountData(it) }
+        data.bitcoinPaymentData?.let { tradeModel.setBitcoinPaymentData(it) }
+        data.paymentProof?.let { tradeModel.setPaymentProof(it) }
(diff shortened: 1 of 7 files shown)


rodvar:
coverage slightly below, accepted for this refactor PR

-------------
Total:   38 lines
Missing: 8 lines
Coverage: 78%
-------------

wodoro:
This is a good refactor, but still far from truly being immutable, thus it does not resolve #1560 imo
To consider it immutable, It will be when there's no setters and all props are also immutable objects, but as mentioned in the issue, its not a perfect world
for now this is good, but I think the problem is with the abstraction, probably with a better abstraction we can have immutability, clean separation and also good performance (maybe a service with granular request for stateflows exposed by this object would be some idea to think about ?)

rodvar:
> This is a good refactor, but still far from truly being immutable, thus it does not resolve #1560 imo To consider it immutable, It will be when there's no setters and all props are also immutable objects, but as mentioned in the issue, its not a perfect world for now this is good, but I think the problem is with the abstraction, probably with a better abstraction we can have immutability, clean separation and also good performance (maybe a service with granular request for stateflows exposed by this object would be some idea to think about ?)

💯  - thanks wod. we are on the same page, except that I do think it resolves the issue/discussion in that issue (title specifically says to remove the tradeState MutableStateFlow usage and this PR does using the _tradeState / tradeState pattern we use in the app)

How about we create a new issue based on this conversation to keep improving in the next months? I'll ask CRAI to do it and please jump in to help tweaking it

@coderabbitai please create a follow up issue to address the concerns described above, might be better to create a discussion first?

PR #1649, opened 2026-07-29
fix(android): disable platform backup, add audited data extraction rules

### LINKED ISSUE #1645: Scope Android allowBackup (dataExtractionRules)
## Motivation

We currently have `apps/clientApp/src/androidMain/AndroidManifest.xml` `android:allowBackup="true"` with **no dataExtractionRules/fullBackupContent**.

### Risk 

 - brief USB/physical access or a malicious restore flow -> adb backup/cloud backup extracts the full app-data dir, including plaintext caches (the encrypted secret survives only as ciphertext).

### What should we do?

 - either set android:allowBackup="false", or add explicit dataExtractionRules/fullBackupContent excluding the DataStore directory.
 - or propose any better solutions if any

wodoro:
## Summary

resolves #1645

Both Android apps were uploading their entire sandbox to Google Drive via Auto Backup.

- `clientApp` had `android:allowBackup="true"` with no rules file (Android Studio template boilerplate).
- `nodeApp` had no `allowBackup` attribute at all, which defaults to `true`.

For `nodeApp` that included `files/Bisq2_mobile/db/private/` — the bisq2 key bundles: identity
private keys, the Tor onion private key, payment accounts, contracts and private trade chat.
For `clientApp` it included the trusted-node `clientSecret` and `sessionId`. Because `clientApp`
has `minSdk 24`, it was also extractable via `adb backup` on API 24–30 without root.

This PR audits every path in both sandboxes and blocks the platform backup pipeline.

## Audit result: nothing is safe to back up

Every candidate fails on confidentiality, correctness, or both.

| Path | Verdict |
|---|---|
| `db/private/` (node) | Identity + onion private keys, payment accounts, trades, private chat |
| `SensitiveSettings.preferences_pb` | Trusted-node bearer credentials (`clientSecret`, `sessionId`) |
| `shared_prefs/bisq_push_notification_keystore.xml` | AES-256 push key, Tink keyset |
| `Settings.preferences_pb` | Device-coupled OS permission mirrors + onboarding gate (see below) |
| `User.preferences_pb` | User-authored content; re-fetched from the node |
| `TradeReadStateMap.preferences_pb` | Leaks trade IDs |
| `OfferbookFilterConfigs.preferences_pb` | Leaks trading interests |
| `ConfigCache` / `BankAccountCountryDetailsCache` | Server caches; stale-restore risk |
| `db/cache/`, `db/network_db/` | Hundreds of MB, fully re-syncable |
| `tor/` | Device/network bound; stale control cookie hangs bootstrap |
| `bisq.log` | Peer onion addresses, trade IDs, profile IDs |
| `bisq_db_from_backup/` | Pending-restore staging dir — see below |
| `Bisq2_mobile/` (client) | Generated CatHash avatars; deterministic |

Three findings worth calling out:

**Keystore-sealed data cannot be restored anyway.** `SensitiveSettings.preferences_pb` and the
push-notification prefs are AES-GCM ciphertext sealed with non-exportable AndroidKeyStore keys.
A restored copy is undecryptable noise on the target device, and the app already treats that as
"keystore invalidated → re-pair". Backing it up buys nothing and leaks the ciphertext.

**`Settings.preferences_pb` looks safe but breaks the install.** It persists
`notificationPermissionState` and `batteryOptimizationState` as mirrors of OS grants.
`DashboardPresenter` and `WelcomeCarouselPresenter` read them straight from the DataStore and never
re-sync from the OS, so a restored `GRANTED` on a device that never granted the permission silently
suppresses the prompt and notifications never arrive. `firstLaunch=false` likewise skips onboarding,
and since the pairing credentials cannot be restored, the user lands unpaired in a post-onboarding
state.

**`bisq_db_from_backup/` is actively dangerous to back up.** It is the staging dir for a pending
user-initiated restore, consumed and deleted on next launch by `maybeRestoreDataDirectory()`.
A snapshot taken between restore and next launch would replay a stale restore over freshly created
keys on the target device.

## Device-to-device transfer

Also excluded. It is a local encrypted channel and materially safer than cloud, but:

- For the client, nothing survives without the Keystore-bound credentials.
- For the node, a silent identity copy risks two live nodes sharing one onion address, and the app
  already ships the correct mechanism: **Settings → Backup**, a user-initiated, optionally
  password-encrypted zip via `NodeBackupServiceFacade`.

## Changes

- `apps/clientApp/src/androidMain/res/xml/data_extraction_rules.xml` — API 31+, full per-file audit
  in comments
- `apps/clientApp/src/androidMain/res/xml/backup_rules.xml` — required because `clientApp` is
  `minSdk 24` and `dataExtractionRules` only applies from API 31
- `apps/nodeApp/src/androidMain/res/xml/data_extraction_rules.xml` — no legacy counterpart;
  `nodeApp` is `minSdk 33`, so a `fullBackupContent` file would be dead config
- Both manifests: `allowBackup="false"` plus the rules wired in as a second layer, so the excludes
  still hold if that attribute is ever relaxed or overridden by a library manifest merge

Rules are deny-by-default per domain rather than per file, so storage added later is opted out until
someone deliberately allows it and records the reasoning. This matters concretely: `android.conf`
has `wallet.enabled = false` today, and when the bisq2 wallet is switched on its seed lands in the
same data dir — covered from day one.

## User-visible impact

Existing users lose Google Drive backup of app data. Given the contents, that is the intended fix.
Recovery paths are unchanged and remain the supported ones:

- **client**: re-pair with the trusted node, which owns the profile and keys
- **node**: Settings → Backup / Restore

## Summary by CodeRabbit

* **Security & Privacy**
  * Disabled automatic Android cloud backups and device-to-device transfers for sensitive app data.
  * Added backup exclusion rules covering app storage, credentials, metadata, and other private data.
  * Applied consistent backup protections across client and node Android apps.

### DIFF
--- a/apps/clientApp/src/androidMain/AndroidManifest.xml
+++ b/apps/clientApp/src/androidMain/AndroidManifest.xml
@@ -17,7 +17,9 @@
 
     <application
         android:name="network.bisq.mobile.client.main.ClientMainApplication"
-        android:allowBackup="true"
+        android:allowBackup="false"
+        android:dataExtractionRules="@xml/data_extraction_rules"
+        android:fullBackupContent="@xml/backup_rules"
         android:enableOnBackInvokedCallback="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
--- a/apps/clientApp/src/androidMain/res/xml/backup_rules.xml
+++ b/apps/clientApp/src/androidMain/res/xml/backup_rules.xml
@@ -18,8 +18,11 @@
   None of it is safe or useful to back up. Recovery on a new device is done by
   re-pairing with the trusted node, which owns the profile and keys.
 
-  Note: android:allowBackup="false" in the manifest already blocks this path.
-  These rules are the second layer, in case that attribute is ever relaxed.
+  Note: on this API range android:allowBackup="false" does block Auto Backup,
+  `adb backup` and D2D, so these rules are a second layer in case that
+  attribute is ever relaxed. That is specific to API 24-30 - from API 31 the
+  attribute no longer opts out of device-to-device transfer, which is why
+  data_extraction_rules.xml carries a load-bearing <device-transfer> section.
 -->
 <full-backup-content>
     <exclude domain="root" path="." />
--- a/apps/clientApp/src/androidMain/res/xml/backup_rules.xml
+++ b/apps/clientApp/src/androidMain/res/xml/backup_rules.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Legacy backup rules for Android 11 and lower (API 24-30).
+
+  Needed because clientApp minSdk is 24: android:dataExtractionRules is only
+  honoured from API 31, so without this file API 24-30 devices would fall back
+  to "back up everything". That window is also the one where `adb backup` can
+  pull the whole sandbox off a USB-debuggable device without root, and where
+  Auto Backup archives are not guaranteed to be encrypted with the device
+  secret - i.e. exactly the devices where a leak would be worst.
+
+  This must mirror the <cloud-backup> section of res/xml/data_extraction_rules.xml.
+  The full per-file audit and the reasoning for the deny-all policy live in that
+  file; it is not duplicated here so the two cannot drift apart.
+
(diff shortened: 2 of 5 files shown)


rodvar (approved):
utACK

Will be tested with next reviews + next release builds

PR #1667, opened 2026-07-30
bugfix/improve_job_cancellation_handling_and_consequent_logging

rodvar:
- fixes done whilst testing #1636 
 - fixed noisy ChildCancelledException logs when loading screens with many cat hashes: superseded offerbook pipeline runs (combined flows) had their cancellations swallowed as reputation failures
 - fixed cancelled pipeline runs wrongly caching valid BUY offers as invalid (cancellation was treated as "no reputation" -> score 0 -> addInvalidBuyOffer)
 - WebSocketApiClient no longer wraps our own cancellation into Result.failure; ensureActive() gate keeps request timeouts (TimeoutCancellationException) reporting as plain failures like before
 - moved try/catch inside withContext in user profile icon facades so the cancellation withContext rethrows isn't logged as an error; genuine cancellations now propagate

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling during profile image loading and WebSocket requests to ensure cancellations propagate appropriately.
  * Preserved fallback profile images when non-cancellation errors occur.
  * Improved trade offer validation to avoid incorrect reputation defaults when operations are cancelled.
* **Tests**
  * Added/extended unit tests covering profile image fallback behavior, WebSocket request cancellation handling, and trade amount limit validation (including cancellation scenarios).

### DIFF
--- a/apps/clientApp/src/androidUnitTest/kotlin/network/bisq/mobile/client/common/domain/service/user_profile/ClientUserProfileServiceFacadeTest.kt
+++ b/apps/clientApp/src/androidUnitTest/kotlin/network/bisq/mobile/client/common/domain/service/user_profile/ClientUserProfileServiceFacadeTest.kt
@@ -1,0 +1,82 @@
+package network.bisq.mobile.client.common.domain.service.user_profile
+
+import io.mockk.every
+import io.mockk.mockk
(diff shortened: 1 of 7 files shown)


rodvar:
merging to include in releases

PR #1668, opened 2026-07-30
refactor(connections): rotate peer card chevron and relocate NetworkInfoDto tests to commonTest

salottodev:
## Summary by CodeRabbit

* **Accessibility**
  * Improved connection card expand/collapse controls with localized screen-reader descriptions.
  * Added a smooth chevron rotation animation to clearly indicate expanded and collapsed states.

* **Tests**
  * Updated UI tests to verify accessibility labels and chevron behavior.
  * Added coverage for decoding connection metrics, including missing and populated metrics data.


apps/clientApp/src/commonTest/kotlin/network/bisq/mobile/client/common/domain/service/network/NetworkInfoDtoTest.kt
```diff
@@ -0,0 +1,66 @@
...
+        val metrics = json.decodeFromString<ConnectionDto>(jsonString).metrics
+
+        // Then
+        assertEquals(184L, metrics?.rttMillis)
+        assertEquals(12_400L, metrics?.sentBytes)
+        assertEquals(340L, metrics?.sentMessageCount)
+        assertEquals(18_900L, metrics?.receivedBytes)
+        assertEquals(512L, metrics?.receivedMessageCount)
+    }
+}
```
rodvar, line 66: please fix: the relocation dropped two of the four cases the old androidUnitTest file had - `ConnectionMetricsDto with an unmeasured rtt decodes rttMillis as null` and `NetworkInfoDto round-trips connections carrying metrics`. moving to commonTest is a win (the kept two now cover the iOS decode path too), but net we lose the null-rtt edge case and the only encode+decode round-trip. both are pure kotlinx.serialization and run fine here - PairingDtoTest/SessionDtoTest already use encodeToString in commonTest - so worth restoring both. 

Is this intentional?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/components/network/ConnectionCard.kt
```diff
@@ -172,11 +182,19 @@ fun ConnectionCard(
...
+                // matching 12dp up-chevron drawable to swap to.
+                ArrowDownIcon(
+                    modifier =
+                        Modifier
+                            .size(12.dp)
+                            .rotate(chevronRotation)
+                            // The asset's own "Down arrow icon" description is meaningless once
+                            // rotated, and rotation is invisible to accessibility services; replace
+                            // it with the localized affordance, as AdvancedOptionsDrawer does.
+                            .clearAndSetSemantics { contentDescription = chevronDescription },
```
rodvar, line 196: nit - the contentDescription sits on the icon while the tap target is the outer card. AdvancedOptionsDrawer (the pattern you're following) puts the description on the clickable element and hides the icon from a11y instead. what you have does satisfy the hint, it's just slightly more disconnected from the click target - fine to leave.

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/design/network/connections/ConnectionsCardDesign.kt
```diff
@@ -538,11 +542,12 @@ internal fun ConnectionCard(
...
-            if (isExpanded) {
-                ArrowDownIcon(modifier = Modifier.size(12.dp))
-            } else {
-                ArrowRightIcon(modifier = Modifier.size(12.dp))
-            }
+            ArrowDownIcon(
+                modifier =
+                    Modifier
+                        .size(12.dp)
+                        .rotate(if (isExpanded) 180f else 0f),
```
rodvar, line 549: nit - the design ref got the rotation but not the clearAndSetSemantics/contentDescription the real ConnectionCard added, so the two diverge on a11y now. it's a kover-excluded POC so no user impact, but since you kept the rotation in sync it'd be tidy to mirror the semantics too.

rodvar (approved):
ACK

💪 

2 nits and one fix for a follow up (I think some tests got lost? - please double check 🙏 )

PR #1712, opened 2026-08-13
chore/improve_trade_funnel_analytics_based_on_first_full_report

### LINKED ISSUE #1711: Trade funnel: instrument cancel/reject reasons + stall time
## Problem

The July 2026 KPI report ([wiki](https://github.com/bisq-network/bisq-mobile/wiki/2026-Monthly-Reports)) shows 208 trades started, ~20% completed. 51% of non-completions are user cancellations or counterparty rejections, and 54 trades sit "still in progress" — but the funnel can't say **why**:

- A liquidity cancel (no response, price moved, wrong payment rail) and a gave-up-on-a-stuck-trade cancel both land in `trade.cancelled`. The second one is an app problem (trade desync / FSM eventQueue bug, bisq-network/bisq2 stuck-trade work) miscounted as a human decision.
- Stuck trades caused by the FSM bug hide in the "in progress" bucket with no signal to separate them from genuinely slow trades.

## Proposal

First improve the analytics to get more insights: Two additions to the sealed trade analytics events (opt-in analytics, privacy contract unchanged — enum values only, no free text, no identity):

1. **Reason enum on `trade.cancelled` / `trade.rejected`** — small sealed set, e.g.:
   - `peer_unresponsive`
   - `price_moved`
   - `payment_method_issue`
   - `no_progress` (user-visible trade state stopped advancing)
   - `changed_mind`
   - `other`

   UX: one optional single-tap chip row on the existing cancel/reject confirmation step — skippable, never free text.

2. **`stall_seconds` on cancel events** — time since the trade's last phase transition, bucketed (e.g. `<1h`, `1–24h`, `1–3d`, `>3d`) rather than raw values. A cancel after days without a transition is a desync fingerprint; minutes is a human decision.

## Why now

- Turns next month's report completion-rate section from a count into a diagnosis (liquidity vs app-caused).
- Once the bisq2 [FSM fix](https://github.com/bisq-network/bisq2/pull/4885) (persist + drain eventQueue) ships, this is the field evidence that it worked: the long-stall cancel bucket and the stale in-progress tail should visibly shrink month over month.

## Scope notes

- Events must be sealed classes per the analytics privacy contract (see `docs/` analytics notes; no per-device identity, opt-in only).
- Reason chips need i18n keys (English base only in `mobile.properties`).
- Phase-transition timestamps: check what the trade model already exposes before adding storage; bucketing happens client-side at event build time.

## Alternative considered

Wait for the bisq2 FSM fix to land and see whether the funnel improves on its own. Downside: without reasons/stall data we can't attribute the improvement, and the liquidity-vs-app question stays open for organic cancels too.

rodvar:
- resolves #1711 
 - added new reasons for interruption of trade analytics
 - bind analytics calls into trade funnel
 - added ui reason prompt to share when cancel/reject trade only visible if the user is opted in to analytics
 - stalled metrics are in-memory only for now. A persisted version follow up will be done if this is proven to be not enough

## Summary by CodeRabbit

* **New Features**
  * Trade cancellation and rejection dialogs can offer selectable reasons, including payment issues, stalled trades, price changes, and changed plans.
  * Trade interruption analytics now record the selected reason and cancellation delay category when analytics is enabled.
  * Dialog actions remain visible while longer content scrolls.
  * Added compact and selected states for chips, with improved visual feedback.

* **Bug Fixes**
  * Trade interruption behavior now respects the user’s analytics preference.
  * Cancellation timing is captured before trade state changes.

### DIFF
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
@@ -141,14 +141,16 @@ class ClientTradesServiceFacade(
         _selectedTrade.value = findOpenTradeItemModel(tradeId)
     }
 
-    override suspend fun rejectTrade(): Result<Unit> {
+    override suspend fun rejectTrade(reason: AnalyticsEvent.Trade.InterruptReason): Result<Unit> {
         if (globalUiManager.notifyIfDemoModeRestricted()) return Result.success(Unit)
-        return apiGateway.rejectTrade(requireNotNull(tradeId)).onSuccess { trackTrade(AnalyticsEvent.Trade.Rejected) }
+        return apiGateway.rejectTrade(requireNotNull(tradeId)).onSuccess { trackTrade(AnalyticsEvent.Trade.Rejected(reason)) }
     }
 
-    override suspend fun cancelTrade(): Result<Unit> {
+    override suspend fun cancelTrade(reason: AnalyticsEvent.Trade.InterruptReason): Result<Unit> {
         if (globalUiManager.notifyIfDemoModeRestricted()) return Result.success(Unit)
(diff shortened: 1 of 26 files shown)


rodvar:
to be included in node release 0.11.0

PR #1734, opened 2026-08-17
feat(privateChat): add two-party private chat with reactions and notifications, reachable from the peer profile

### LINKED ISSUE #590: Chat: Private chats outside of trade
Image

Image

Image

## Designs

 - Please see designs composable previews and comments already uploaded
 - Note it absorbs Messages-tab activation + DM notifs; reuses #545 + #589.

## Desktop Reference

Image

Reuse Trade chat UI

We can make this accesible only via finding the profile avatar and clicking on chat.
The chat persists interaction with that peer.

Once the community hub is implemented it should be accessible from the private chats tab as per designs

salottodev:
depends on: https://github.com/bisq-network/bisq2/pull/4940
resolves: https://github.com/bisq-network/bisq-mobile/issues/590

### Issue found: Deep Links and Notification Taps on a Cold Start Land on a Not-Yet-Connected App

Found while smoke-testing this branch.

With the app killed, `adb shell am start -a android.intent.action.VIEW -d "bisq://PrivateChat/<id>"` — and equally a notification tap, which builds the same `ACTION_VIEW` intent — opens the target screen **on top of the still-bootstrapping splash**. Pressing back returns to `ClientSplashScreen`, still connecting.

This is not specific to private chat, and this PR does not introduce it:

-   `NavigationManagerImpl.navigateFromUri` is route-agnostic, and the root `NavHost` starts at `NavRoute.Splash()` (`ClientRootNavGraph:47`, `NodeNavGraph:37`), so `OpenTrade` and `TradeChat` behave the same way.
-   `OpenTradePresenter:107-116` is in fact less forgiving: it reads a synchronous snapshot and raises the trade-not-found dialog immediately, with no grace period at all. `PrivateChatPresenter` waits `CHANNEL_WAIT_TIMEOUT_MS = 5_000` first.

What makes it more visible here is that a DM's primary entry point is a push notification tap with the app closed. Not reproduced yet, but plausible: if connecting takes longer than 5s — Tor on the node build, websocket on Connect — a channel that does exist could render as "this conversation is no longer available".

Sketch, if we want to fix it separately: gate inside `navigateFromUri` — wait for the root destination to leave `Splash`, then navigate only if it settled on `TabContainer`, and drop the pending URI otherwise, since without a profile there is nothing to open. `isAtMainScreen()` already has that check, and the `tabDestinationListener` in the same file is the precedent for observing destinations. It would also mean the 5s budget is counted from "app is on the main screen" rather than from a cold start.

Happy to open a separate issue for this.


rodvar:
we decided to split this in 2 PRs (one for node and one for connect)

also the described issue has been created to address separately 🙏

PR #1748, opened 2026-08-18
fix: don't report cancelled trade steps as failures (#1706)

### LINKED ISSUE #1706: Coroutine cancellation reported as a trade-step failure
When a trade-step action is cancelled (e.g. the user navigates away mid-request), the resulting `CancellationException` ("Job was cancelled") surfaces as a `TradeActionException: Trade step '...' request failed` and is captured as an error. Cancellation is normal control flow, not a failure — it shouldn't be reported to the user as a failed step nor logged as an analytics error.

- Project: **Bisq Connect (Android)**
- Release: `bisq-connect@0.7.1`
- Occurrences: **3** across two steps
  - `btc_received` request failed: Job was cancelled — http://uzlqr4bnoscfifvjvzj3kicoc23ejibefpsynuorcvetjptzqkaqisid.onion/bisq/issues/138
  - `btc_address` request failed: Job was cancelled — http://uzlqr4bnoscfifvjvzj3kicoc23ejibefpsynuorcvetjptzqkaqisid.onion/bisq/issues/130

**tech comment**: In the trade-step action path, rethrow `CancellationException` ahead of the broad `catch` (do not wrap it in `TradeActionException`, do not surface a "request failed" message, do not capture it as analytics). Only genuine failures should produce the error state + event.

testable as navigating away of a trade

nostrbuddha:
- Fixes https://github.com/bisq-network/bisq-mobile/issues/1706
 - Tried hardware back during `btc address send` step. Issue not reproducible. The global loading overlay swallows `back navigate`, so the presenter never cancels.
 - The 0.7.1 reports (with same guardedSuspendAction() gate) probably came from other teardown (process death, rotation, etc.,)
 - `TradeAnalyticsTracker.trackAction` now uses `ensureActive()` before reporting, same gate as `WebSocketApiClient`.
    - If this screen’s job is already cancelled, skip `FAILED` / GlitchTip.
    - If the user is still on the confirm and the request times out, still report `FAILED`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cancelled trade actions.
  * Active request timeouts are recorded as failed analytics events.
  * Caller-initiated cancellations now stop processing without creating unnecessary analytics reports or captured exceptions.

### DIFF
--- a/shared/domain/src/androidUnitTest/kotlin/network/bisq/mobile/domain/service/trades/TradeAnalyticsTrackerTest.kt
+++ b/shared/domain/src/androidUnitTest/kotlin/network/bisq/mobile/domain/service/trades/TradeAnalyticsTrackerTest.kt
@@ -3,10 +3,14 @@ package network.bisq.mobile.domain.service.trades
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.verify
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.advanceTimeBy
 import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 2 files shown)


rodvar:
@coderabbitai could you please create a follow up issue considering this PR description, I would summarise it as : "the global loading overlay swallows back-navigate, so the presenter can never cancel even if user wants to go back"

rodvar (approved):
utACK 

will be tested in next reviews
elegant well tested fix - thanks buddha!

rodvar:
I keept thinking and I think this won't fully resolve the issue although it helps..

please consider a follow up analysing these:

1) make the gate an unconditional ensureActive() at the top of onFailure (covers non-CE teardown failures too)
2) the node facade's catch (Exception) { Result.failure(e) } wrappers are the root cause and also leak "StandaloneCoroutine was cancelled" into user-facing snackbars — a shared gated wrapper there fixes analytics AND UI at once
3) the second test can pass vacuously if the UNDISPATCHED start ever gets dropped — add a positive assert that the CE actually propagated, and fix the comment saying the node facade wraps teardown into Result.failure (it doesn't; withContext rethrows at its boundary). Also watch the 1706 reports after next release — they should drop but maybe not to zero until (2).

let me know if you have any questions/thoughts - thanks @nostrbuddha

nostrbuddha:
Thanks for the comments @rodvar 

Agree to these. I ll do a follow PR.

PR #1765, opened 2026-08-21
bugfix/repairing_to_new_trusted_node_should_wipe_backend_capabilities_data_as_well

rodvar:
- implemented sensitive backend data wipe when user deicides to switch to a different node (even if he ends up in the same one). This fixes backend capabilities gating update bug. implemented using TDD
 - found whilst testing private chat feature

## Summary by CodeRabbit

* **New Features**
  * Configuration now reacts to client ID changes during device pairing.
  * Updated configuration data is fetched automatically after re-pairing.
  * Stale configuration loads are cancelled when connection settings change.

* **Bug Fixes**
  * Prevented unnecessary cache clearing and refetching when unrelated settings are updated.
  * Preserved cached configuration during startup.


rodvar:
merged to continue testing, any feedback will be addressed as follow up

PR #1767, opened 2026-08-21
refactor(push): wrap key with Android Keystore

### LINKED ISSUE #1764: Migrate Android push-notification key storage off deprecated `androidx.security-crypto`
## Background

`shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/crypto/PushNotificationKey.android.kt` stores the AES-256 symmetric key that the trusted node uses to encrypt push payloads (decrypted in `BisqFirebaseMessagingService`). It uses `MasterKey` + `EncryptedSharedPreferences` from `androidx.security-crypto`, which Google deprecated wholesale in 1.1.0. The APIs still work — this produces 9 deprecation warnings per build today, nothing is broken — but the library is unmaintained going forward and should be replaced deliberately rather than under pressure later. There is already an in-code TODO at the top of the file pointing at direct Keystore APIs.

## Requested approach: direct Android Keystore wrap

Generate an AES-GCM key in `AndroidKeyStore` and use it to wrap/unwrap the symmetric key; store the wrapped blob + IV in plain SharedPreferences.

Why this over the industry-consensus Tink + Jetpack DataStore combo:

- The store holds exactly **one 32-byte key** under a single pref key (`bisq_push_notification_keystore` / `push_notification_symmetric_key_base64`). Tink earns its weight with keysets, rotation policies and multi-key management — none of which apply here. Direct Keystore is the smallest sound solution and adds zero new dependencies.
- The current implementation deliberately persists with synchronous `commit()` because the key is registered with the trusted node immediately after the write; if the persist were async and the process died, node and device would disagree on the key and pushes would silently fail to decrypt. Plain SharedPreferences keeps that write-before-register guarantee trivially, while DataStore is async-first and would need `runBlocking` gymnastics to preserve it.
- What we give up: Tink's armor against OEM Keystore quirks (odd `KeyStoreException`s on some devices, key invalidation after lock-screen changes). The implementation must handle Keystore failures defensively — treat an unwrap failure as "no key" and let the normal rotation path recover, never crash the messaging service.

## What makes this easier than it looks

- `getOrCreatePushNotificationKeyBase64()` **rotates the key on every call** (fresh key, overwrite, re-register with the node) — mirroring the iOS Keychain behaviour. There is no long-lived dataset to preserve, only the currently registered key.
- A test seam already exists (`PushNotificationKeyStore` interface + `pushNotificationKeyStoreFactory`) because Robolectric cannot emulate `AndroidKeyStore`. The migration must keep this seam.
- iOS is untouched (Keychain via `PushNotificationKeyStore.swift`); this is `androidMain` only. Production call sites are clientApp-only: registration in `ClientPushNotificationServiceFacade`, decrypt in `BisqFirebaseMessagingService`.

## Research first (decides the migration story)

- [ ] **When does key rotation actually run?** If device registration (and therefore rotation + node re-registration) happens on every app start / opt-in refresh, the old stored key never needs to be carried over: first launch on the new version rotates into the new store and the node gets the new key — "migration" is then just deleting the legacy prefs file. If registration is rarer (only at pairing / opt-in), a one-time read-old → write-new carry-over is needed to avoid a window of undecryptable pushes.
- [ ] Confirm no other data lives in `bisq_push_notification_keystore` besides the one key.

## Acceptance criteria

- [ ] No `androidx.security-crypto` imports remain; the 9 deprecation warnings are gone; no new dependencies added.
- [ ] Key remains hardware-backed (Android Keystore) at rest; never stored in plaintext.
- [ ] The write-then-register ordering guarantee is preserved (synchronous persistence before the key is sent to the trusted node).
- [ ] Keystore failures degrade gracefully: unwrap/decrypt errors surface as "no key" and recover via the normal rotation path; `BisqFirebaseMessagingService` never crashes on a bad store.
- [ ] `PushNotificationKeyStore` seam and in-memory fake still work for unit tests in other modules.
- [ ] Legacy `bisq_push_notification_keystore` prefs file is cleaned up after migration.
- [ ] New storage file(s) covered by the existing backup-exclusion / data-extraction rules (platform backups are disabled with audited rules — verify the new file names are within the audited set).
- [ ] Tested on: fresh install, upgrade-with-existing-key (push decryption keeps working across the upgrade, or is proven unnecessary per the research item), and Robolectric unit tests.

wodoro:
androidx.security-crypto was deprecated wholesale in 1.1.0, so replace EncryptedSharedPreferences with an AES-GCM key in AndroidKeyStore that wraps the symmetric key, stored as Base64(iv || ciphertext) in plain SharedPreferences. This reuses LocalEncryption under its own alias, so there is no new crypto and no new dependency; the library is dropped from both modules and from the version catalog.

The old store is deleted rather than carried over. The key rotates on every registration and the first activate() after the upgrade re-registers a fresh one, so only pushes arriving before the next app launch are affected, and reading the old store would mean keeping the dependency this change removes.

Keystore failures degrade to "no key" so the messaging service drops the push instead of crashing, and the existing PushNotificationKeyStore test seam is preserved. Keystore-only code is excluded from Kover, since Robolectric ships no AndroidKeyStore provider; it is covered by an instrumented test instead.

resolves #1764

## Summary by CodeRabbit

* **Security Improvements**
  * Push-notification keys are protected with Android Keystore-backed encryption before local storage.
  * Key access recovers safely after corruption or unavailable encryption keys.

* **Bug Fixes**
  * Push notifications now register automatically when enabled and a user profile becomes available.
  * Duplicate device registrations are prevented.

* **Maintenance**
  * Legacy push-notification encryption data is removed during migration.
  * Backup and restore documentation reflects the updated key-storage behavior.

* **Testing**
  * Added coverage for key protection, recovery, registration, and legacy-data cleanup.

### DIFF
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
@@ -10,6 +10,8 @@ import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import network.bisq.mobile.client.common.domain.sensitive_settings.SensitiveSettingsRepository
 import network.bisq.mobile.data.crypto.getOrCreatePushNotificationKeyBase64
@@ -57,6 +59,16 @@ class ClientPushNotificationServiceFacade(
 
     private val _deviceId = MutableStateFlow<String?>(null)
     private val deviceId: StateFlow<String?> = _deviceId.asStateFlow()
+
+    // Registering is not idempotent: every run rotates the symmetric key and sends the new one
+    // to the trusted node. Two overlapping runs can interleave as rotate(k1), rotate(k2),
+    // register(k1), register(k2), leaving the node with a key the device no longer holds, which
+    // silently breaks decryption of every push until the next registration. Opting out has the
+    // same problem in reverse: a registration landing after it would leave the device registered
+    // at the node while the user believes they are off. The triggers are independent
+    // (auto-registration, Settings toggle, Support screen, FCM token refresh), so serialize them
+    // rather than assume they never overlap.
+    private val registrationMutex = Mutex()
 
     override suspend fun activate() {
         super<ServiceFacade>.activate()
@@ -174,7 +186,9 @@ class ClientPushNotificationServiceFacade(
             Result.failure(PushNotificationException("Device identifier temporarily unavailable", e))
         }
 
-    private suspend fun registerTokenWithTrustedNode(token: String): Result<Unit> {
+    private suspend fun registerTokenWithTrustedNode(token: String): Result<Unit> = registrationMutex.withLock { registerTokenWithTrustedNodeLocked(token) }
+
+    private suspend fun registerTokenWithTrustedNodeLocked(token: String): Result<Unit> {
         // Get the current user profile (needed for publicKeyBase64)
         val userProfile = userProfileServiceFacade.selectedUserProfile.value
         if (userProfile == null) {
@@ -232,7 +246,9 @@ class ClientPushNotificationServiceFacade(
         return result
     }
 
-    override suspend fun unregisterFromPushNotifications(): Result<Unit> {
+    override suspend fun unregisterFromPushNotifications(): Result<Unit> = registrationMutex.withLock { unregisterFromPushNotificationsLocked() }
+
+    private suspend fun unregisterFromPushNotificationsLocked(): Result<Unit> {
         log.i { "Unregistering from push notifications..." }
 
         // deviceId tells the server which device to unregister. On iOS it can be transiently
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
@@ -1,5 +1,6 @@
 package network.bisq.mobile.client.common.domain.service.push_notification
 
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -114,6 +115,11 @@ class ClientPushNotificationServiceFacade(
             } else {
                 log.w { "Auto-registration failed: ${result.exceptionOrNull()?.message}" }
             }
+        } catch (e: CancellationException) {
+            // deactivate() cancels serviceScope while this may be suspended in fetch() or the
+            // registration network call. CancellationException is an Exception on JVM, so the
+            // handler below would swallow it and report a shutdown as a registration error.
+            throw e
         } catch (e: Exception) {
             log.e(e) { "Error during auto-registration" }
         }
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
@@ -5,6 +5,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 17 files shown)


wodoro:
https://github.com/bisq-network/bisq-mobile/wiki/Push-Notifications-Android

will need an update after this PR is merged probably (EncryptedSharedPreferences)

wodoro:
note: I tried testing the notification relay on main branch first so then try with this PR afterwards, but I didnt have any chance using either of the implementations (notifications didn't arrive)
so I appreciate if you can test it to make sure things work

(I have created a firebase app and downloaded the json file with .debug suffix and put it in the project files, and it enables successfully in the app without any errors)

rodvar:
> note: I tried testing the notification relay on main branch first so then try with this PR afterwards, but I didnt have any chance using either of the implementations (notifications didn't arrive) so I appreciate if you can test it to make sure things work
> 
> (I have created a firebase app and downloaded the json file with .debug suffix and put it in the project files, and it enables successfully in the app without any errors)

I'll ping you on matrix about this

rodvar:
rebased to main

shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/crypto/LegacyPushNotificationKeyCleanup.kt
```diff
@@ -0,0 +1,85 @@
...
+private const val LEGACY_PREFS_FILE = "bisq_push_notification_keystore"
+private const val LEGACY_MASTER_KEY_ALIAS = "_androidx_security_master_key_"
+
+private val legacyStoreCleaned = AtomicBoolean(false)
+
+/**
+ * Drops the `EncryptedSharedPreferences` store used before the Keystore-wrap
+ * migration. The old key is deliberately not carried over: every registration
+ * rotates the key anyway, and the first `activate()` after the upgrade
+ * re-registers a fresh one with the trusted node. Pushes that arrive in between
```
rodvar, line 28: please fix: manual upgrade testing shows this assumption doesn't hold, and it's the linchpin of the whole migration.

What actually happens on the first launch after upgrade: auto-registration fires from the settings collect in `ClientPushNotificationServiceFacade.activate()` before `selectedUserProfile` is populated (profile data arrives from the trusted node over Tor seconds later), so `registerDevice` aborts with "Cannot register device: no user profile selected" — before it ever reaches key rotation. And nothing retries: the settings trigger only fires on the enabled-flag transition, and the other registration paths are manual (Settings toggle, Support screen) or an FCM token refresh. Since key rotation is also what constructs `SharedPrefsKeyStore` and runs this cleanup, the migration never executes at all. Verified on emulator: after upgrade the legacy `bisq_push_notification_keystore.xml` is still in `shared_prefs/`, no `bisq_push_notification_key.xml` was created, and the log shows `Auto-registration failed: No user profile selected`.

The race predates this PR, but in 0.8.2 it's benign — a failed re-registration doesn't matter while the old key is still readable. With this PR the old key becomes unreadable while the node keeps encrypting with it, so upgrading push users get every push silently dropped until they manually re-toggle notifications. The only out-of-PR mitigation would be keeping the legacy read fallback this PR exists to remove, so the fix belongs here: make auto-registration wait for / re-trigger on `selectedUserProfile` becoming non-null (e.g. combine the settings flow with `selectedUserProfile` in `activate()`, register when both conditions hold, guarded by `_isDeviceRegistered`).

shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/crypto/PushNotificationKey.android.kt
```diff
@@ -73,41 +64,75 @@ actual fun getOrCreatePushNotificationKeyBase64(): String? =
...
- * (i.e. the user has not opted in / registered for push notifications).
+ * (i.e. the user has not opted in / registered for push notifications) or if the
+ * stored blob can no longer be unwrapped. A Keystore that lost or invalidated the
+ * wrapping key has to degrade to "no key" (drop the push) rather than crash the
+ * messaging service. The next registration rotates into a fresh, usable key.
  */
 fun readPushNotificationKeyBase64(): String? =
     runCatching {
         pushNotificationKeyStoreFactory().get()
     }.getOrNull()
```
rodvar, line 87: please fix: this runCatching (and the one in `getOrCreatePushNotificationKeyBase64` above) swallows the exception without logging, so the messaging service logs the same "No push notification symmetric key on device" whether there was never a key or the Keystore just blew up. OEM keystore quirks are exactly the thing we'd need logs for in the field — an `.onFailure { log.e(it) { ... } }` before `getOrNull()` would do. These exceptions carry no key material, so they're safe to log.

shared/domain/src/androidUnitTest/kotlin/network/bisq/mobile/data/crypto/LegacyPushNotificationKeyCleanupTest.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+/**
+ * Deletes with the migration it covers (see `LegacyPushNotificationKeyCleanup`).
+ * The Keystore half of the cleanup, dropping the orphaned master key, only runs on
+ * a device and is covered by `PushNotificationKeyStoreInstrumentedTest`.
```
rodvar: nit - this says the Keystore half is covered by `PushNotificationKeyStoreInstrumentedTest`, but `legacyEncryptedSharedPrefsStoreIsDeleted` there only asserts the prefs entry is gone — it never checks that the `_androidx_security_master_key_` alias actually got deleted, nor that it survives when there was no legacy file (the case the production guard exists for). Either extend the instrumented test or soften this claim.

shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/crypto/LegacyPushNotificationKeyCleanup.kt
rodvar: this is a hidden preexisting race condition that was harmless before but we need to fix now please 🙏

rodvar (changes requested):
Very solid implementation code wise. However it seems it needs a bit more work to ensure it works as expected.

**shared prefs before upgrade**

Image

**after**

Image

**should have removed the bisq_push_notification_keystore.xml in favour of the new bisq_push_notification_key.xml** ..?


Please see my inline comments, the most important issue is the long paragraph one

shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/crypto/LegacyPushNotificationKeyCleanup.kt
wodoro: I thought we agreed on letting the pushes silently get dropped in that gap?
from what I understand we just need to fix the race condition?

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/ClientPushNotificationServiceFacade.kt
```diff
@@ -63,17 +65,32 @@ class ClientPushNotificationServiceFacade(
...
+                _isPushNotificationsEnabled,
+                userProfileServiceFacade.selectedUserProfile,
+            ) { pushEnabled, userProfile ->
+                pushEnabled && userProfile != null
+            }.distinctUntilChanged()
+                .collect { canAutoRegister ->
+                    if (canAutoRegister && !_isDeviceRegistered.value) {
+                        tryAutoRegisterIfOnboarded()
+                    }
                 }
```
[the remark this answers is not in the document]
wodoro: Both getOrCreatePushNotificationKeyBase64 and readPushNotificationKeyBase64 are plain non-suspend functions, and everything inside their runCatching blocks is blocking work: SecureRandom, Base64 encoding, SharedPreferences.commit(), and Keystore calls. There is no suspension point, so no CancellationException can originate inside them. The caller wraps the call in withContext(backgroundDispatcher), and cancellation surfaces there, outside the runCatching.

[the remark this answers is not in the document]
wodoro: so does [ba82639](https://github.com/bisq-network/bisq-mobile/pull/1767/commits/ba826397b517b00b22368ef8fc170e5c96933c67) resolve this ?

shared/domain/src/androidMain/kotlin/network/bisq/mobile/data/crypto/LegacyPushNotificationKeyCleanup.kt
rodvar: discussed on matrix - race condition deemed to be fixed in last commit (its not related to the missing first push after migration which we will accept exceptionally for this change)

rodvar:
resuming reviewing on this one, rebased to main

rodvar (approved):
ACK

all concerns have been addressed + it works exactly as agreed/discussed (we accept a first message loss after device migration to eliminate the deprecated dependecy)

thanks @wodoro

rodvar:
there is a small cleanup to do, include proguard rules after this - I'll send a follow up in a sec and would appreciate your review on it (should be small PR)

PR #1800, opened 2026-09-01
fix: skip GlitchTip capture for expected protocol-validation rejections

### LINKED ISSUE #1707: Message validation rejections captured as analytics errors
`TradeProtocolException`s that represent bisq2 core **correctly rejecting a bad or hostile peer message** are being captured as error-level events in GlitchTip. These are working-as-designed protective rejections, not app faults — they add noise to the error stream and can mask real crashes.

- Project: **Bisq Easy Node (Android)**, release `bisq-easy-node@0.10.0`
- Examples:
  - `Bitcoin address length must not be longer than 62` — thrown by `BisqEasyBtcAddressMessageHandler.verify` (over-length BTC address from peer). 2026-08-05 — http://uzlqr4bnoscfifvjvzj3kicoc23ejibefpsynuorcvetjptzqkaqisid.onion/bisq/issues/148
  - `Takers (buyers) Bitcoin amount is too high...` — thrown by `BisqEasyTakeOfferRequestHandler.validateAmount` (amount mismatch / possible price manipulation). 2026-08-04 — http://uzlqr4bnoscfifvjvzj3kicoc23ejibefpsynuorcvetjptzqkaqisid.onion/bisq/issues/147

### Proposed solutions

1. Classify protocol-validation rejections as expected outcomes and **do not** send them as error events (preferred), or
2. Downgrade them to a distinct, low-severity signal (e.g. `trade.peer_message_rejected`) with **no** peer-supplied content in the payload — useful to measure "how often do we receive malformed/hostile trade messages" without noise or PII.

nostrbuddha:
- Fixes https://github.com/bisq-network/bisq-mobile/issues/1707
 - Classifies known bisq2 protocol-validation rejections (Better solution should be to have named exceptions in bisq2 and work based on them).
 - Prevents expected peer-message rejections from being captured as GlitchTip exceptions.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved trade error handling for expected validation rejections.
  - Trade errors continue to be shown and recorded, while known protocol-related rejections are no longer treated as unexpected exceptions.
  - Unexpected trade failures continue to be captured for diagnostics, improving the accuracy of error reporting and analytics.

### DIFF
--- a/shared/domain/src/androidUnitTest/kotlin/network/bisq/mobile/domain/service/trades/TradeAnalyticsTrackerTest.kt
+++ b/shared/domain/src/androidUnitTest/kotlin/network/bisq/mobile/domain/service/trades/TradeAnalyticsTrackerTest.kt
@@ -228,7 +228,7 @@ class TradeAnalyticsTrackerTest {
         }
 
     @Test
-    fun `observeTrades emits Errored and captures the exception on a local error`() =
+    fun `observeTrades emits Errored and captures unexpected local errors`() =
         runTest {
             val analytics = mockk<AnalyticsService>(relaxed = true)
             val scope = CoroutineScope(UnconfinedTestDispatcher(testScheduler))
@@ -245,7 +245,7 @@ class TradeAnalyticsTrackerTest {
         }
 
     @Test
-    fun `observeTrades emits Errored on a peer-only error even when the local error stays null`() =
+    fun `observeTrades emits Errored on a peer-only unexpected error even when the local error stays null`() =
         runTest {
             val analytics = mockk<AnalyticsService>(relaxed = true)
             val scope = CoroutineScope(UnconfinedTestDispatcher(testScheduler))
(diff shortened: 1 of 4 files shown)


rodvar (approved):
utACK

Verified all 7 prefixes verbatim against bisq2 core and confirmed the FSM stores the raw root-cause message so startsWith matches what's persisted. 

only a small nit on the PR: I think the "rather than peer-supplied text" comment overclaims: `peersErrorMessage` does come over the wire; worth noting the mismatch fails open to capture instead.

 Also heads up for the follow-up: bisq2 already has TradeProtocolFailure with an `isUnexpected` flag covering 3 of these 7 cases -> the typed fix is plumbing that through the DTO plus adding enum entries for the address/invoice validations, rather than new named exceptions. Shall I create an issue to address this in the near future?

nostrbuddha:
Yes sure. And I will update the comment (the nit) in the next PR.

PR #1814, opened 2026-09-03
build: move app packaging to AGP 9 APIs and enable ABI splits

### LINKED ISSUE #569: Apk splitting per architecture
For the `.apk` releases in node, we could get a significant downsize in the apk if we apply architecture apk splitting techniques.

This is quite complex for an app like node (with the kmp, tor ,etc deps) hence this issue to study how would we do it and configure gradle for smooth generation

Gains for Connect at the time of writing are not significant so doesn't worth the hassle.

wodoro:
closes https://github.com/bisq-network/bisq-mobile/issues/569

Size effect on release: Bisq Connect drops from 40.0 MB universal to 22.1 MB for arm64-v8a; Bisq Easy from 96.9 MB to 79.0 MB (smaller proportional win, since the bisq2 dex dominates there).

a surprise that I discovered during this investigation: Official node app release is bundling a macOS specific tor artifact, adding 18.6 MB unnecessarily to the package
I have applied a fix on this PR for it too, but it needs checking on mac builds. this is why the official release is 114MB and not the 96.9 MB reported here

notes:
- From next release after this PR, the version codes will jump from 32 and 50 to 32000 and 50000 for example, because the abi splitting needs predictable and dedicated version codes for clean distribution (32000 being universal and 32001, 32002, 32003, 32004 being the split apks)
- I included the connect app, because the numbers show it is significant enough that splitting is worth it in my opinion
- Also during the same investigation for MacOS binaries, found out that the apk includes netty native binaries for desktop OS, which is around 9.6MB uncompresed, the PR includes that fix too (~3.3 MB compressed, so Bisq Easy down to ~ 75 MB)
- I intentionally tried to avoid deprecated APIs that are removed in AGP 9, because a migration is in sight (#1682), hence the introduction of build-logic

## Summary by CodeRabbit

* **New Features**
  * Android builds now provide separate APKs for supported CPU architectures, along with a universal APK.
  * APK naming and versioning are standardized across app variants.
  * Release signing checks now verify every generated APK split.

* **Bug Fixes**
  * Android packages no longer include desktop-only or unsupported native binaries.

* **Documentation**
  * Updated Android packaging and verification instructions for split and universal APKs.

* **Chores**
  * Improved build formatting checks and centralized ktlint version management.

### DIFF
--- a/apps/nodeApp/build.gradle.kts
+++ b/apps/nodeApp/build.gradle.kts
@@ -211,6 +211,13 @@ android {
             excludes.add("META-INF/INDEX.LIST")
             excludes.add("META-INF/NOTICE.markdown")
 
+            // bisq2's JVM tor stack contributes a desktop tor bundle as a jar resource. It only
+            // materializes when the build host is macOS, which is how the published APKs are
+            // produced: Bisq_Easy-0.12.0.apk carries an 18.6 MB tor.zip of Mach-O binaries
+            // (tor, libevent dylib, lyrebird, conjure-client) that cannot execute on Android.
+            // Android drives tor through kmp-tor's lib/<abi>/libtor.so instead, so drop it.
+            excludes.add("tor.zip")
+
             pickFirsts.add("**/protobuf/**/*.class")
             pickFirsts +=
                 listOf(
--- a/apps/nodeApp/build.gradle.kts
+++ b/apps/nodeApp/build.gradle.kts
@@ -25,6 +25,7 @@ plugins {
     alias(libs.plugins.kotlin.serialization)
     alias(libs.plugins.protobuf)
     alias(libs.plugins.kover)
+    id("network.bisq.mobile.app-artifacts")
 }
 
 // -------------------- Version Configuration --------------------
@@ -98,6 +99,11 @@ val releaseKeystoreFile: File? = extra["releaseKeystoreFile"] as File?
 @Suppress("UNCHECKED_CAST")
 val optionalSigningProp = extra["optionalSigningProp"] as (String) -> String
 extra["requiredCompanionProps"] = listOf("KEYSTORE_PASSWORD", "KEY_ALIAS", "KEY_PASSWORD")
+
+// -------------------- App artifact naming / packaging --------------------
+appArtifacts {
+    productName.set(appName)
+}
 
 // -------------------- Android Configuration --------------------
 android {
@@ -181,10 +187,16 @@ android {
         }
     }
 
-    // Disable ABI splits to avoid packaging conflicts with kmp-tor
+    // ABI splits: one APK per architecture instead of one carrying all four copies of
+    // libtor.so. The universal APK stays for sideloading, and AppArtifactsPlugin gives each output
+    // its own version code. That plugin also switches this off for bundle builds, which AGP
+    // cannot produce while splits are on.
     splits {
         abi {
-            isEnable = false
+            isEnable = true
+            reset()
+            include("armeabi-v7a", "arm64-v8a", "x86", "x86_64")
+            isUniversalApk = true
         }
     }
 
@@ -275,15 +287,6 @@ android {
             matchingFallbacks += listOf("release")
         }
     }
-    applicationVariants.all {
-        val variant = this
-        outputs.all {
-            val output = this as com.android.build.gradle.internal.api.BaseVariantOutputImpl
-            val version = variant.versionName
-            val fileName = "${appName.replace(" ", "_")}-$version.apk"
-            output.outputFileName = fileName
-        }
-    }
     buildFeatures {
         buildConfig = true
     }
@@ -297,9 +300,6 @@ android {
     testOptions {
         unitTests.isIncludeAndroidResources = true
     }
-
-    // Needed for aab files renaming
-    setProperty("archivesBaseName", getArtifactName(defaultConfig))
 }
 
 // -------------------- Protobuf Configuration --------------------
@@ -518,6 +518,3 @@ afterEvaluate {
             }
     }
 }
-
-// -------------------- Helper Functions --------------------
-fun getArtifactName(defaultConfig: com.android.build.gradle.internal.dsl.DefaultConfig): String = "${appName.replace(" ", "")}-${defaultConfig.versionName}_${defaultConfig.versionCode}"
--- a/apps/nodeApp/build.gradle.kts
+++ b/apps/nodeApp/build.gradle.kts
@@ -218,12 +218,14 @@ android {
             // Android drives tor through kmp-tor's lib/<abi>/libtor.so instead, so drop it.
             excludes.add("tor.zip")
 
-            // grpc-netty-shaded ships its tcnative binaries as jar resources for every desktop
-            // OS. Windows .dll and macOS .jnilib cannot be loaded on Android by any code path,
-            // and no Linux or Android tcnative build is present, so netty already falls back to
-            // the JDK TLS stack here. 9.6 MB of formats the platform cannot read.
-            excludes.add("META-INF/native/*.dll")
-            excludes.add("META-INF/native/*.jnilib")
+            // grpc-netty-shaded ships tcnative and epoll binaries as jar resources for every
+            // desktop OS: Windows .dll, macOS .jnilib and glibc-linked Linux .so. None can be
+            // loaded on Android. AGP's java-resource merger already drops the .so on its own, so
+            // only the .dll and .jnilib (9.6 MB) actually reached the APK, but excluding the whole
+            // directory covers all of them and does not rot if that merger behaviour changes.
+            // Nothing under META-INF/native/ is loadable here: Android takes native code from
+            // lib/<abi>/ only, so netty stays on the JDK TLS stack and NIO transport either way.
+            excludes.add("META-INF/native/**")
 
             pickFirsts.add("**/protobuf/**/*.class")
             pickFirsts +=
--- a/apps/nodeApp/build.gradle.kts
+++ b/apps/nodeApp/build.gradle.kts
@@ -218,6 +218,13 @@ android {
             // Android drives tor through kmp-tor's lib/<abi>/libtor.so instead, so drop it.
             excludes.add("tor.zip")
 
+            // grpc-netty-shaded ships its tcnative binaries as jar resources for every desktop
+            // OS. Windows .dll and macOS .jnilib cannot be loaded on Android by any code path,
+            // and no Linux or Android tcnative build is present, so netty already falls back to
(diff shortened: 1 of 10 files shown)


rodvar:
@wodoro wod thanks so much for addressing this. It looks like a big win. Before I start reviewing and testing I've got a few questions kindly respond when you have a moment:

 1. I don't understand why we need that logic on version codes. A person installing one arch won't need the other hence it makes sense that every release build has the same version code regardless of the arch it is for, no? Also, this splitting doesn't affect stores (where there are rules for version codes) as what goes in there is the bundle (.aab) and they do their own splitting. Plus, why the big loss in version codes (32 -> 32000 ?) is this really needed? 
 2. are the std gradle tasks used for testing?  anything I need to know? did you smoked tests the builds (at least check no crush when starting?) 
 
 Thanks in advance 🙏

wodoro:
1. The version codes for different APKs become necessary once we try to publish this as split APKs to F-droid. it has math for creating version codes, and each needs different version code as far as I understand/remember. (see https://f-droid.org/docs/Build_Metadata_Reference/#VercodeOperation). otherwise this wouldn't be necessary

2. I have run the APKs on the emulator and they did not crash, and Bisq Easy connected to tor and showed offers, but beyond that it needs more testing

apps/nodeApp/build.gradle.kts
```diff
@@ -199,6 +211,20 @@ android {
...
+            // (tor, libevent dylib, lyrebird, conjure-client) that cannot execute on Android.
+            // Android drives tor through kmp-tor's lib/<abi>/libtor.so instead, so drop it.
+            excludes.add("tor.zip")
+
+            // grpc-netty-shaded ships its tcnative binaries as jar resources for every desktop
+            // OS. Windows .dll and macOS .jnilib cannot be loaded on Android by any code path,
+            // and no Linux or Android tcnative build is present, so netty already falls back to
+            // the JDK TLS stack here. 9.6 MB of formats the platform cannot read.
+            excludes.add("META-INF/native/*.dll")
+            excludes.add("META-INF/native/*.jnilib")
```
rodvar: please fix: grpc-netty-shaded 1.81.0 also ships linux tcnative + epoll `.so` files under `META-INF/native/` (~6.1 MB uncompressed - I've checked the jar in my gradle cache: `libio_grpc_netty_shaded_netty_tcnative_linux_x86_64.so` 3.03 MB, `linux_aarch_64.so` 2.81 MB, plus two epoll `.so`). They're glibc-linked so Android can't load them either, but the `*.dll`/`*.jnilib` excludes leave them in - so about half the dead weight this commit targets is still in the APK, and the comment saying "no Linux or Android tcnative build is present" would mislead the next person?

Suggest also excluding `META-INF/native/*.so` (or just the whole `META-INF/native/` dir) and adjusting the comment, waiting your review 🙏

apps/nodeApp/build.gradle.kts
```diff
@@ -181,10 +187,16 @@ android {
...
 
-    // Disable ABI splits to avoid packaging conflicts with kmp-tor
+    // ABI splits: one APK per architecture instead of one carrying all four copies of
+    // libtor.so. The universal APK stays for sideloading, and AppArtifactsPlugin gives each output
+    // its own version code. That plugin also switches this off for bundle builds, which AGP
+    // cannot produce while splits are on.
     splits {
         abi {
-            isEnable = false
+            isEnable = true
```
rodvar, line 196: The comment this replaces said splits were disabled "to avoid packaging conflicts with kmp-tor" — do we know what that conflict actually was? it was added alongside the useLegacyPackaging workaround for kmp-tor's exec resources, which is still in place. The size numbers prove packaging, but not that tor bootstraps from a per-ABI apk on a real device — and under the new version codes the arm64 split is the artifact most users will end up with. Before this ships I'd want a tor bootstrap smoke with the arm64-v8a split on a device (you confirmed you did, I will do to before merging 💪 )

build-logic/src/main/kotlin/network/bisq/gradle/AppArtifactsPlugin.kt
```diff
@@ -0,0 +1,187 @@
...
+            baseVersionCode: Int,
+            abi: String?,
+        ): Int = baseVersionCode * ABI_VERSION_CODE_MULTIPLIER + (ABI_ORDINALS[abi] ?: 0)
+    }
+}
+
+// Splits are a DSL flag, decided long before the task graph exists, so the requested task names
+// are the only signal available. Gradle keys configuration cache entries on them, so branching
+// here stays cache-correct.
+private fun Project.isBundleRequested(): Boolean =
```
rodvar: this one is more of a question and we might do changes:

I generally run for example for node release: `./gradlew apps:nodeApp:clean apps:nodeApp:assembleRelease apps:nodeApp:bundleRelease` to get a clean all-builds (till now, universal) in bundle and apk. 

How would that play now? I would expect it to generate the exact same .aab and .apk as before this PR + all the per-arch ones. If that's the case then great. What if I want a specific arch only? Let's work on leaving a clear solution here that doesn't break release process 💪

build-logic/src/main/kotlin/network/bisq/gradle/AppArtifactsPlugin.kt
```diff
@@ -0,0 +1,187 @@
...
+                // the splits rather than the shrinker.
+                if (dsl.splits.abi.isEnable && project.isBundleRequested()) {
+                    dsl.splits.abi.isEnable = false
+                    project.logger.lifecycle(
+                        "ABI splits disabled for ${project.path}: incompatible with bundle tasks.",
+                    )
+                }
+            }
+
+            androidComponents.onVariants { variant ->
```
rodvar, line 63: please fix: splits + the rename transform apply to every variant, so a plain `assembleDebug` on nodeApp now packages 5 unshrunken apks (~100 MB each) and RenameApksTask byte-copies all of them again - roughly a GB of extra packaging/IO per inner-loop build where only one device ABI matters.

Could we please implement gating splits (or at least the rename wiring) to release variants, and/or hard-linking instead of `copyTo`?
For debug default should be pre-PR behaviour (with universal) unless a param is specified?

rodvar (changes requested):
amazing work Wod, this also give us part of the needed work we need to do for the AGP toolchain migration :D 

I did a first code-only pass, after your review and edits I'll do a full pass including testing all cases. Thank you

apps/nodeApp/build.gradle.kts
wodoro: AGP strips `.so` so it never reached APK even if it was in the jar
but will exclude the META-INF/native entirely to prevent accidental future regressions

build-logic/src/main/kotlin/network/bisq/gradle/AppArtifactsPlugin.kt
wodoro: now assemble debug produces a universal APK, like before
and the APK produced is hard linked instead of copied, and falls back to copy if hardlinking unavailable, so IO is reduced significantly

abi splits are now gated and can be controlled: `-PabiSplits=true|false` and `-Pabi=arm64-v8a[,x86_64]` (comma separated or only one)

build-logic/src/main/kotlin/network/bisq/gradle/AppArtifactsPlugin.kt
wodoro: > ABI splits and app bundles cannot be built in one invocation (AGP fails with
  "Multiple shrunk-resources files found", https://issuetracker.google.com/402800800).
  Run them separately, for :apps:nodeApp:
    ./gradlew :apps:nodeApp:bundleRelease
    ./gradlew :apps:nodeApp:assembleRelease
  or pass -PabiSplits=false to build both at once with a single universal APK.

So: two invocations gets you the same .aab and a universal .apk as before plus the per-arch ones; `-PabiSplits=false` keeps your single command with exactly pre-PR output

wodoro: However I suggest building all 5 APKs for release moving forward, 1 universal APK + 4 variants (or 3 variants since x86 is probably useless)
we will need them for f-droid reproducible builds

wodoro: actually thinking about this again, I guess for f-droid we need a bit more preparations, I'll send another PR after this is merged, explaining what we need to do going forward
a quick summary would be: Dockerfile based build artifacts from CI -> build, compare for reproducibility down to bytes and sign locally -> upload signed artifacts to GitHub release
(we can also skip the build and compare locally when f-droid pipeline is ready, which does this for us, but we will find out later when it doesn't get released on f-droid)

build-logic/src/main/kotlin/network/bisq/gradle/AppArtifactsPlugin.kt
rodvar: excellent

build-logic/src/main/kotlin/network/bisq/gradle/AppArtifactsPlugin.kt
rodvar: I'm not sure I understood the last parenthesis properly, I'll ping you on that one on chat

For the rest, you mean we will have a new CI-based signatured just for F-Droid releases hold by our GH? This might be great so the new store distribution wouldn't add any overload once we get it right.

rodvar:
the one liner I used to apply to get the builds

 - `./gradlew apps:[type]:clean apps:[type]:assembleRelease apps:[type]:bundleRelease --info`

should now. be relplaced with:

 - `./gradlew apps:[type]:clean apps:[type]:bundleRelease --info && ./gradlew apps:[type]:assembleRelease --info` 
 
 where [type] = clientApp|nodeApp

rodvar:
rebased to main + added release management docs to the android.md file

rodvar (approved):
ACK

 - all concerns have been addressed
 - tested on real and virtual devices on all cases
 - iOS fully unnafected as expected
 - this PR shaves `~17.5%` of the universal builds, optimisation might be higher for the aab based one from google play. Adding it to the original optimizations done that sums up a total of closer to 50% reduction (!). Shave is even higher if user opts to download its specific architecture (which is probably what happens under the hood in google play as well)
 - as a plus also already tackles needed build changes in our path to #1682 
 - there is a residual risk that will be tested on each respective release: I'm not sure how the big jump in versionCodes will affect Google Play these days. Technically shouldn't matter but is yet to be seen. I'll make any necessary adjustments if needed on each next release.

Awesome work 💪 🚀

PR #1821, opened 2026-09-04
fix: hold cold-start deep links until the app reaches the main screen

### LINKED ISSUE #1729: Deep links and notification taps on cold start navigate on top of the still-connecting Splash screen
## Context

The following applies to `relayed Push notifications` setups (iOS + opt-in connect Android).

Seems to be work missing from the push notifications implementation.

With the app **killed**, opening a deep link — or tapping a push notification, which builds the same `ACTION_VIEW` intent — navigates to the target screen **immediately on top of the still-bootstrapping Splash**. The app hasn't connected yet (Tor on the node build, websocket on Connect), so the target presenter runs against empty state, and pressing back returns to the Splash screen, still connecting.

## Acceptance criteria

- Cold-start deep link / notification tap shows Splash normally and navigates to the target only after the app reaches the main screen; back from the target goes to the main screen, not Splash.
- A deep link to an existing trade/conversation no longer shows "not found / no longer available" just because connecting took longer than the presenter's grace period.
- If startup lands on onboarding instead of the main screen, the pending URI is discarded (no navigation fired into onboarding).
- Warm-start deep links (app already on main screen) behave exactly as today.

wodoro:
Fixes #1729.

## The problem

With the app killed, opening a deep link, or tapping a push notification which builds the same `ACTION_VIEW` intent, navigated to the target screen immediately, on top of the still-bootstrapping splash. The app had not connected yet (Tor on the node build, the websocket on Connect), so the target presenter ran against empty state, and pressing back returned to the splash, still connecting.

This is not specific to any one route. `NavigationManagerImpl.navigateFromUri` is route-agnostic and the root `NavHost` starts at `NavRoute.Splash`, so `OpenTrade`, `TradeChat` and `PrivateChat` all behaved the same way.

### 1. The navigation gate

`navigateFromUri` holds a link that arrives while the root destination is still `Splash`, and navigates only once startup settles on `TabContainer`. If startup lands anywhere else, such as the user agreement, onboarding, profile creation or the trusted-node setup, the link is dropped, because without a profile there is nothing to open. A newer link supersedes an older pending one. Warm starts return immediately, so a link that arrives with the app already up behaves exactly as before.

**The wait is unbounded, on purpose.** There is nothing to time out against. The user is looking at a splash they cannot navigate away from, so the pending link has no competing intent to go stale against, and a startup that never finishes ends in a restart that takes the coroutine with it. A bound could only ever discard a link that startup was still going to honour, which is the same silent-drop failure this PR exists to fix. The cost of waiting is one suspended coroutine holding a string.

The wait follows the current controller through `rootNavControllerFlow` rather than capturing the one that was current when the link arrived. Startup can replace the controller when the activity is recreated, and a captured controller stops emitting, which would leave the link waiting forever and doing nothing. The controller used for the actual navigation is fetched after the wait, for the same reason.

If the controller cannot report where startup is at all, the link is dropped rather than fired blind, since navigating blind is the failure this PR is fixing.

### 2. Trade and chat data: a sync signal rather than a grace period

Reaching the main screen means bootstrap completed, but on the Connect build the trades and their chat messages can still be arriving. Both trade presenters read a synchronous snapshot of them, and `OpenTradePresenter` raised its trade-not-found dialog immediately, with no grace period at all.

A fixed grace period cannot solve this, because it is wrong in both directions. Too short and the UI lies: on a cold start, trade chat messages land around ten seconds after the screen opens. Too long and a trade that really is gone stalls before saying so. The underlying problem is that an empty collection means "not delivered yet" and "empty" at the same time, and only the data layer knows which.

So both facades publish the distinction:

- `TradesServiceFacade.openTradesSynced`, backed in `BaseTradesServiceFacade`. The client sets it on the `ModificationType.REPLACE` snapshot it already keys on. `WebSocketClientImpl.subscribe` synthesises that event on every successful subscribe, so it is guaranteed to arrive. The node sets it at the end of `activate()`, where the bisq2 `CollectionObserver`s have already replayed the existing collection synchronously while binding.
- `TradeChatMessagesServiceFacade.chatMessagesSynced`, set by the client after the first trade-chat payload is applied to the channels, and by the node in `activate()` for the same reason.

`selectOpenTradeWhenSynced` waits on the first of those. Either the trade resolves, or the trades sync without it, in which case it really is gone. An existing trade is never reported as no longer available, and a stale notification for a closed trade says so at once instead of after a fixed stall. The lookup goes through the facade rather than a local `find`, so the client's short-trade-id fallback keeps working, and it runs in the `first { }` predicate rather than the `combine` transform, which can be called for values that never reach the decision.

### 3. Loading state on the trade screens

Both trade screens rendered a blank body during those waits, which was fine when the lookup was synchronous. They now render `LoadingState` inside the scaffold, so the top bar and its back button stay live, and in trade chat so does the message input. This follows the precedent `PrivateChatScreen` already set.

`OpenTradeScreen` derives its loading state from the trade and the not-found dialog. `TradeChatPresenter` carries an explicit `isLoading` flag instead, because its messages arrive separately and an empty list on its own cannot be told apart from a chat that has not loaded. The flag is driven by `chatMessagesSynced`.

## Notes on the shape of the change

`initialize` is not a suspend function on either presenter. Each launches into `presenterScope` and keeps a `tradeJob` it cancels on re-initialisation, resetting the state that describes the previous trade along with it, which is what `PrivateChatPresenter` does. Keeping it non-suspend leaves cancellation with the presenter rather than splitting it between the composition scope and `presenterScope`.

## Acceptance criteria from #1729

- [x] Cold-start deep link or notification tap shows the splash normally and navigates to the target only after the app reaches the main screen; back from the target goes to the main screen, not the splash.
- [x] A deep link to an existing trade or conversation no longer shows "not found / no longer available" because connecting took longer than a grace period.
- [x] If startup lands on onboarding instead of the main screen, the pending URI is discarded.
- [x] Warm-start deep links behave exactly as today.

## Testing

Unit tests pass across `shared:domain`, `shared:presentation`, `apps:clientApp` and `apps:nodeApp`; ktlint clean. Diff coverage against `upstream/main` is **90%**, above the project's 80% threshold (`./gradlew koverPRCheck`).

New coverage: the cold-start gate (waits then navigates, discards on onboarding, does not fire while startup stays on the splash, last link wins, drops when the splash state cannot be determined), `selectOpenTradeWhenSynced` (resolves immediately, resolves on a later sync update, reports absent once synced, keeps waiting while unsynced), the trade chat loading flag, the trade-not-found path, and the client's `REPLACE` snapshot setting `openTradesSynced` while an incremental update does not.

What is still uncovered, for transparency: the two Compose screens' loading branches, which are UI code, and the ignored-user message filter in both trade presenters, which this PR only re-indented. Nothing covers the tab-graph deep-link path after a cold start, only the root-graph path, and the node's `activate()` setting `openTradesSynced` is argued from the bisq2 observer semantics rather than asserted by a test. The iOS targets were not built. The changed code is `commonMain` and the client facade is shared, so iOS takes the same path, but that is reasoning rather than verification.

Verified on device for the reported scenario: notification tap with the app killed.

## Summary by CodeRabbit

* **New Features**
  * Added loading indicators while trade details and trade chats synchronize.
  * Trade and chat views now wait for initial data before showing “not found” or empty states.
  * Added synchronization status and failure handling for open trades and trade messages.
  * Deep links received during the splash screen are queued until startup reaches the main screen.
  * Improved trade selection when data arrives during synchronization.

* **Bug Fixes**
  * Prevented stale trade or chat content from appearing during re-initialization.
  * Improved handling of missing trades and superseded startup deep links.
  * Prevented sending chat messages before a trade is available.

### DIFF
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/trade/ClientTradeChatMessagesServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/trade/ClientTradeChatMessagesServiceFacade.kt
@@ -4,6 +4,7 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
@@ -40,6 +41,9 @@ class ClientTradeChatMessagesServiceFacade(
     private val allChatReactions: MutableStateFlow<Set<BisqEasyOpenTradeMessageReaction>> =
         MutableStateFlow(emptySet())
 
+    private val _chatMessagesSynced = MutableStateFlow(false)
+    override val chatMessagesSynced: StateFlow<Boolean> = _chatMessagesSynced.asStateFlow()
+
     // Misc
     override suspend fun activate() {
         super<ServiceFacade>.activate()
@@ -68,6 +72,7 @@ class ClientTradeChatMessagesServiceFacade(
     }
 
     override suspend fun deactivate() {
+        _chatMessagesSynced.value = false
         super<ServiceFacade>.deactivate()
     }
 
@@ -82,6 +87,9 @@ class ClientTradeChatMessagesServiceFacade(
             updatedTradeIds.forEach { tradeId ->
                 updateChatMessages(tradeId)
             }
+            // Set after the channels are updated: the first payload is the node's snapshot, so from
+            // here a channel with no messages is empty rather than still loading.
+            _chatMessagesSynced.value = true
         }
     }
 
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/trade/ClientTradeChatMessagesServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/trade/ClientTradeChatMessagesServiceFacade.kt
@@ -2,6 +2,8 @@ package network.bisq.mobile.client.common.domain.service.chat.trade
 
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
@@ -10,6 +12,8 @@ import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.serialization.json.Json
 import network.bisq.mobile.client.common.domain.util.notifyIfDemoModeRestricted
+import network.bisq.mobile.client.common.domain.websocket.WebSocketClientService
+import network.bisq.mobile.client.common.domain.websocket.subscription.Topic
 import network.bisq.mobile.client.common.domain.websocket.subscription.collectPayloads
 import network.bisq.mobile.data.replicated.chat.Citation
 import network.bisq.mobile.data.replicated.chat.reactions.BisqEasyOpenTradeMessageReaction
@@ -22,6 +26,7 @@ import network.bisq.mobile.data.service.chat.trade.TradeChatMessagesServiceFacad
 import network.bisq.mobile.data.service.trades.TradesServiceFacade
 import network.bisq.mobile.data.service.user_profile.UserProfileServiceFacade
 import network.bisq.mobile.presentation.common.ui.base.GlobalUiManager
+import kotlin.concurrent.Volatile
 
 class ClientTradeChatMessagesServiceFacade(
     private val tradesServiceFacade: TradesServiceFacade,
@@ -29,6 +34,7 @@ class ClientTradeChatMessagesServiceFacade(
     private val apiGateway: TradeChatMessagesApiGateway,
     private val json: Json,
     private val globalUiManager: GlobalUiManager,
+    private val webSocketClientService: WebSocketClientService,
 ) : ServiceFacade(),
     TradeChatMessagesServiceFacade {
     // Properties
@@ -41,8 +47,19 @@ class ClientTradeChatMessagesServiceFacade(
     private val allChatReactions: MutableStateFlow<Set<BisqEasyOpenTradeMessageReaction>> =
         MutableStateFlow(emptySet())
 
+    // The first TRADE_CHAT_MESSAGES payload is the node's snapshot, but applying it needs the user
+    // profile as well: updateChatMessages skips every channel until the profile lands, which on a cold
+    // start over Tor can be after the snapshot. Synced is the two together, re-evaluated by whichever
+    // arrives second.
+    @Volatile
+    private var chatMessagesSnapshotReceived = false
     private val _chatMessagesSynced = MutableStateFlow(false)
     override val chatMessagesSynced: StateFlow<Boolean> = _chatMessagesSynced.asStateFlow()
+
+    private val _chatMessagesSyncFailed = MutableStateFlow(false)
+    override val chatMessagesSyncFailed: StateFlow<Boolean> = _chatMessagesSyncFailed.asStateFlow()
+
+    private var tradeChatsJob: Job? = null
 
     // Misc
     override suspend fun activate() {
@@ -61,19 +78,34 @@ class ClientTradeChatMessagesServiceFacade(
                 if (tradeId != null) {
                     updateChatMessages(tradeId = tradeId)
                 }
+                updateChatMessagesSynced()
             }
         }
-        serviceScope.launch {
-            subscribeTradeChats()
-        }
+        tradeChatsJob =
+            serviceScope.launch {
+                subscribeTradeChats()
+            }
         serviceScope.launch {
             subscribeChatReactions()
+        }
+        // A subscribe that failed is only retried on the next reconnect, so until then the snapshot is
+        // not coming and a wait on it has to be told rather than left spinning.
+        serviceScope.launch {
+            webSocketClientService.failedSubscriptionTopics.collect { failed ->
+                _chatMessagesSyncFailed.value = Topic.TRADE_CHAT_MESSAGES in failed
+            }
         }
     }
 
     override suspend fun deactivate() {
+        // Joined, not just cancelled: a payload the collector is still applying would otherwise flip
+        // the flags back after the reset, on a facade that is gone.
+        tradeChatsJob?.cancelAndJoin()
+        tradeChatsJob = null
+        chatMessagesSnapshotReceived = false
         _chatMessagesSynced.value = false
         super<ServiceFacade>.deactivate()
+        _chatMessagesSyncFailed.value = false
     }
 
     private suspend fun subscribeTradeChats() {
@@ -87,10 +119,15 @@ class ClientTradeChatMessagesServiceFacade(
             updatedTradeIds.forEach { tradeId ->
                 updateChatMessages(tradeId)
             }
-            // Set after the channels are updated: the first payload is the node's snapshot, so from
-            // here a channel with no messages is empty rather than still loading.
-            _chatMessagesSynced.value = true
+            // After the channels are updated, so a channel with no messages is empty rather than still
+            // loading by the time synced reads true.
+            chatMessagesSnapshotReceived = true
+            updateChatMessagesSynced()
         }
+    }
+
+    private fun updateChatMessagesSynced() {
+        _chatMessagesSynced.value = chatMessagesSnapshotReceived && selectedUserProfileId.value != null
     }
 
     private suspend fun subscribeChatReactions() {
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
@@ -1,5 +1,6 @@
 package network.bisq.mobile.client.common.domain.service.trades
 
+import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -106,6 +107,7 @@ class ClientTradesServiceFacade(
     }
 
     override suspend fun deactivate() {
+        setOpenTradesSynced(false)
         openTradesSubscription.dispose()
         closedTradesSubscription.dispose()
         tradePropertiesSubscription.dispose()
@@ -255,7 +257,10 @@ class ClientTradesServiceFacade(
             }
 
     // Private
-    private fun handleTradeItemPresentationChange(
+
+    /** internal (not private) to expose the snapshot/increment handling as a same-module unit-test seam. */
+    @VisibleForTesting
+    internal fun handleTradeItemPresentationChange(
         payload: List<TradeItemPresentationDto>,
         modificationType: ModificationType,
     ) {
@@ -267,6 +272,8 @@ class ClientTradesServiceFacade(
                     applyPendingTradeProperties(tradeModel)
                 }
                 _openTradeItems.value = newTrades
+                // The snapshot is the whole truth, so from here a missing trade is a trade that is gone.
+                setOpenTradesSynced(true)
             }
 
             ModificationType.ADDED -> {
--- a/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
+++ b/apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
@@ -6,6 +6,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
 import kotlinx.serialization.json.Json
 import network.bisq.mobile.client.common.data.mapping.trade.toClosedTradeListItem
 import network.bisq.mobile.client.common.domain.util.notifyIfDemoModeRestricted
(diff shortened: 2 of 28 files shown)


shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/navigation/manager/NavigationManagerImpl.kt
```diff
@@ -211,49 +218,116 @@ class NavigationManagerImpl(
...
+                .first { !it.destination.hasRoute<NavRoute.Splash>() }
+
+        if (!pendingDeepLinkUri.compareAndSet(uri, null)) {
+            log.i { "Dropping deep link ${uri.deepLinkRoute()}, superseded by a newer one" }
+            return false
+        }
+
+        val reachedMainScreen = settled.destination.hasRoute<NavRoute.TabContainer>()
+        if (!reachedMainScreen) {
+            log.i { "Dropping deep link ${uri.deepLinkRoute()}, startup landed on ${settled.destination.route} instead of the main screen" }
```
rodvar: question - a user who already has a profile but cold-starts into a new user-agreement version also lands off the main screen, so their link gets dropped silently here. the "without a profile there is nothing to open" rationale doesn't quite cover that case. probably fine vs the issue's AC, but maybe worth a note or a small follow-up issue?

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/trades/TradesServiceFacade.kt
```diff
@@ -73,3 +82,26 @@ interface TradesServiceFacade : LifeCycleAware {
...
 }
+
+/**
+ * Selects [tradeId] and returns it, retrying while the open trades sync in. A deep link or a
+ * notification tap opens a trade right after the app connects, when the list can still be arriving,
+ * so a plain snapshot read reports a trade that does exist as missing. Returns null once the trades
+ * have synced without it, which means genuinely absent and is what the callers' not-found dialog is
+ * for.
+ */
+suspend fun TradesServiceFacade.selectOpenTradeWhenSynced(tradeId: String): TradeItemPresentationModel? {
```
rodvar, line 108: the first {} predicate mutates shared facade state (selectOpenTrade / selectedTrade) on every emission, and two overlapping waits for different trade ids race on that global: deep link opens trade A, a chat notification for trade B arrives before the snapshot lands, both waiters run selectOpenTrade when it does (null on every miss, last writer wins). send/cancel/reject all read the same global, so the visible screen's action can target the other trade. worth resolving the trade locally instead of through the shared selection

apps/nodeApp/src/androidMain/kotlin/network/bisq/mobile/node/common/domain/service/trades/NodeTradesServiceFacade.kt
```diff
@@ -180,9 +180,14 @@ class NodeTradesServiceFacade(
                     }
                 },
             )
+
+        // Both observers replay the existing collection while binding, so the open trades are complete
+        // by the time activate() returns.
+        setOpenTradesSynced(true)
```
rodvar, line 186: the synced-at-activate contract (here and in NodeTradeChatMessagesServiceFacade) rests on bisq2 CollectionObservers replaying synchronously during bind - you flag this yourself in the PR description as argued rather than tested. if that ever changes upstream this silently regresses to the old race. could we add a small node-side test asserting `openTradesSynced` is true right after activate() to pin this down please?

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
```diff
@@ -255,7 +257,10 @@ class ClientTradesServiceFacade(
             }
 
     // Private
-    private fun handleTradeItemPresentationChange(
+
+    /** internal (not private) to expose the snapshot/increment handling as a same-module unit-test seam. */
+    @VisibleForTesting
```
rodvar, line 278: nit - internal + @VisibleForTesting purely as a test seam is fine here, but let's keep an eye on these so they don't accumulate?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/navigation/manager/NavigationManagerImpl.kt
```diff
@@ -211,49 +218,116 @@ class NavigationManagerImpl(
...
+                .filterNotNull()
+                .flatMapLatest { it.currentBackStackEntryFlow }
+                .first { !it.destination.hasRoute<NavRoute.Splash>() }
+
+        if (!pendingDeepLinkUri.compareAndSet(uri, null)) {
+            log.i { "Dropping deep link ${uri.deepLinkRoute()}, superseded by a newer one" }
+            return false
+        }
+
+        val reachedMainScreen = settled.destination.hasRoute<NavRoute.TabContainer>()
```
rodvar: question - dropping unless the FIRST non-splash destination is TabContainer also discards links across transient interstitials that do continue to the main screen: TrustedNodeSetup(showConnectionFailed=true) after a flaky connect (user taps retry and reaches main seconds later, tapped trade never opens), and user-agreement re-acceptance for an existing user (splash checks tac before hasUserProfile), which contradicts the 'without a profile there is nothing to open' rationale. worth handling these two or at least a follow-up issue

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/navigation/manager/NavigationManagerImpl.kt
```diff
@@ -211,49 +218,116 @@ class NavigationManagerImpl(
...
+                // Navigating blind could stack the target on a splash that is not ready; a link that is
+                // not opened is the lesser harm.
+                log.w { "Dropping deep link ${uri.deepLinkRoute()}, cannot tell whether startup is still on the splash" }
+                return false
+            }
+            true -> Unit
+        }
+
+        log.i { "Holding deep link ${uri.deepLinkRoute()} until startup leaves the splash" }
+        pendingDeepLinkUri.value = uri
```
rodvar: please fix: the supersede write happens before any hasDeepLink validation, so a second splash-time uri that matches no nav graph (MainActivity forwards any dataString with no pre-filter) still kills a valid held link and then no-ops itself - nothing navigates.  alidate before superseding, or move the write after?

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/navigation/manager/NavigationManagerImpl.kt
```diff
@@ -211,49 +218,116 @@ class NavigationManagerImpl(
...
+        // Follows the current controller rather than capturing one: a controller replaced during
+        // startup stops emitting, and a captured one would leave the link waiting forever.
+        val settled =
+            rootNavControllerFlow
+                .filterNotNull()
+                .flatMapLatest { it.currentBackStackEntryFlow }
+                .first { !it.destination.hasRoute<NavRoute.Splash>() }
+
+        if (!pendingDeepLinkUri.compareAndSet(uri, null)) {
```
rodvar: a bit of an edge case but worth noting: last-one-wins only covers links that were both held: a link arriving right after the splash settles takes the warm path without touching pendingDeepLinkUri, so an older held link's CAS still succeeds and it navigates on top of the newer one. a replaceable Job for the waiter (cancel the previous on supersede) would fix this, the issue above, and stop superseded waiters leaking

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/trade/trade_detail/OpenTradePresenter.kt
```diff
@@ -109,62 +111,81 @@ class OpenTradePresenter(
...
+        _isTradeOutOfSync.value = false
+        // The chat row renders as soon as the trade is published, which happens before its collector
+        // runs, so these would carry the previous trade's numbers into the new one.
+        msgCount.value = 0
+        _lastChatMsg.value = null
+
+        tradeJob?.cancel()
+        tradeJob =
+            presenterScope.launch {
+                val currentTrade = tradesServiceFacade.selectOpenTradeWhenSynced(tradeId)
```
rodvar, line 142: please fix: this wait has no error path - on the client, if the TRADES subscribe fails once (WebSocketClientService logs 'Failed to subscribe... skipping' and only retries on the next reconnect) openTradesSynced never flips and the screen spins forever. before this PR the user at least got the actionable not-found dialog. needs a retry or an error escape hatch (same wait in TradeChatPresenter.initialize)

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/trade/trade_detail/OpenTradePresenter.kt
```diff
@@ -109,62 +111,81 @@ class OpenTradePresenter(
...
 
-        val currentTrade = _selectedTrade.value
-        if (currentTrade == null) {
-            log.w { "OpenTradePresenter.initialize called but selectedTrade is null - skipping flow collection" }
-            _showTradeNotFoundDialog.value = true
-            return
-        }
+        // Reset with the job: re-initialising resolves another trade, and until it does, the previous
+        // trade's details and its not-found dialog do not describe what the screen is showing.
+        _showTradeNotFoundDialog.value = false
```
rodvar, line 126: nit - this reset block misses _tradeAbortedBoxVisible/_tradeProcessBoxVisible (only reset in onViewUnattaching), so a singleTop re-navigate from a rejected trade A to trade B flashes A's interrupted pane on B until B's state collector fires (barely noticeable in fast devices but worth fixing)

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/trade/trade_chat/TradeChatScreen.kt
```diff
@@ -69,7 +71,12 @@ fun TradeChatScreen(tradeId: String) {
             )
         },
     ) {
-        if (readCount == -1) {
+        if (isLoading) {
```
rodvar, line 74: the input stays enabled while loading, but sendChatMessage does require(selectedTrade.value != null) - typing and sending before the spinner clears throws into the coroutine exception handler and the message is dropped. Maybe gate send on !isLoading or on the presenter's resolved trade?

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/chat/trade/ClientTradeChatMessagesServiceFacade.kt
```diff
@@ -82,6 +87,9 @@ class ClientTradeChatMessagesServiceFacade(
             updatedTradeIds.forEach { tradeId ->
                 updateChatMessages(tradeId)
             }
+            // Set after the channels are updated: the first payload is the node's snapshot, so from
+            // here a channel with no messages is empty rather than still loading.
+            _chatMessagesSynced.value = true
```
rodvar: Haven't tested this yet, but I think there might be two problems with flipping synced here: 

 - (1) updateChatMessages early-returns while selectedUserProfileId is still null, so on a cold start over tor this claims synced with every channel still empty and the chat renders wrongly empty until the profile lands. 
 - (2) the TRADE_CHAT_MESSAGES subscription is gated on openTradeItems being non-empty plus a subscribe that, once failed, is skipped until the next reconnect - then synced stays false all session and the chat spins forever (pre-PR it rendered an empty but usable chat)

apps/clientApp/src/commonMain/kotlin/network/bisq/mobile/client/common/domain/service/trades/ClientTradesServiceFacade.kt
```diff
@@ -106,6 +107,7 @@ class ClientTradesServiceFacade(
     }
 
     override suspend fun deactivate() {
+        setOpenTradesSynced(false)
```
rodvar: please fix: synced is reset before disposing the subscription, and the handler runs on Dispatchers.Default - an in-flight REPLACE can flip it back to true on a deactivated facade. also unlike the node facade, _openTradeItems is never cleared here (nor on activate), so after re-pairing a deep link can short-circuit on a previous session's trade or see a false not-found. same ordering in ClientTradeChatMessagesServiceFacade.deactivate

rodvar (changes requested):
very solid PR wod, did a first code-only pass will fully test early next week. Hope this helps 🙏

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/navigation/manager/NavigationManagerImpl.kt
wodoro: the suggested fix changes what "last one wins" means. If B is the user's most recent tap and it cannot open, they land on A, a screen they did not just ask for, with no hint that B failed. A tap that opens nothing is at least honest.

rodvar: I agree with what you are saying wod, but that doesn't address my point: MainActivity forwards any dataString unfiltered, so B might not be an ask at all — malformed/foreign uris shouldn't get to kill a held link. and graph validation can't run pre-supersede anyway (tab controller doesn't exist at splash)

So how about this:

 1. do a cheap scheme/parse to remove any noise before the pending write and
 2. UX improvement nit - a snackbar when a superseding link then can't open would make the "honest no-op"visible instead of silent :)

rodvar:
rebased to main

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/common/ui/navigation/manager/NavigationManagerImpl.kt
```diff
@@ -231,49 +243,133 @@ class NavigationManagerImpl(
...
+        }.onFailure { e ->
+            log.e(e) { "Failed to determine if at splash (nav graph may not be ready yet)" }
+        }.getOrNull()
+
+    /**
+     * Deep links reach [navigateFromUri] from outside the app and carry ids (a trade, a channel, a
+     * profile) after the route. Diagnostics only need to know which route was held or dropped, so the
+     * id never reaches the log.
+     */
+    private fun String.deepLinkRoute(): String = substringAfter("://").substringBefore('/')
```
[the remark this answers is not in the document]
rodvar: nit - deepLinkRoute leaks the whole string to the log when the uri has no "://" (and query ids for foreign uris) — same unfiltered-dataString class as my supersede comment. since both bite on exactly the same input, maybe one stone: a scheme/parse check at navigateFromUri's mouth kills the supersede edge AND means only well-formed uris ever reach the logger; or just harden the logger with substringAfter("://", "") + cut at /?# if you'd rather keep supersede as-is.

I believe this is related to our discussion above. I've also found a bug whilst testing the solution I'll report in a bit, maybe you can address both in a follow up? let me know wod 🙏 

@wodoro

rodvar (approved):
ACK (Android)

Most concerns fully addresed.

iOS will be fully tested pre-release by the end of the week on an internal release build

please check latest CRAI comment and [this issue](https://github.com/bisq-network/bisq-mobile/issues/1829) which would be nice to fix ideally by tomorrow 🙏 

awesome work as usual 💪

PR #1842, opened 2026-09-10
bugfix/fix_push_notifications_key_pair_rotation_too_fast_causing_notif_loss_in_certain_cases

### LINKED ISSUE #1835: Opted-in trade push notifications being lost across sessions (pairing key re-registration issue)
found during #1831 testing, using an opted-in for relayed notifications Android device.

**steps**

 1. start a new trade to buy btc, send the btc address and kill the app
 2. on the peer device, send the fiat payment info
 3. before the notification arrive open the app until you arrive to dashboard, then kill again (this regenerates key pair for PN)
 4. wait for the push notification

ER: should arrive in a few seconds
OR: never arrives

Potential solutions/issues to fix:

- Tolerate one key generation of skew: keep the previous key alongside the current one and try both on decrypt - one-file change in the key store + messaging service, kills the whole class of failure. Alternatively rotate only when no key exists or on explicit re-pair, but the two-key window is safer for queued FCM messages either way.
- Stale registration pruning: your node still holds the dead d35CM… token from an old emulator image (deviceId 49423c2a96fcd7b2) - every dispatch burns a relay call that FCM rejects NotRegistered, and nothing feeds that back to unregister it. Same for the old bisq2_api_headless dir's ancient tokens.
- Privacy nit: WebSocketClientImpl debug-logs the full registration body - device token and symmetricKeyBase64 - into logcat (that's how I read your key). onNewToken carefully refuses to log even a token prefix; the WS request log bypasses all of that. Worth verifying it's stripped in release builds and redacting the key regardless.

rodvar:
- resolves #1835
 - avoided immediately discarding previous relay push notifications key pair when a new one is registered, to be able to decrypt messages sent using the previous related public key. Implementation provided for both android and ios + tests
 - cleanup bisq2 side still needed (trusted node), created separate issue for it

## Summary by CodeRabbit

- **New Features**
  - Push notifications now remain decryptable across one symmetric-key rotation on Android and iOS.
  - The app retains the previous notification key temporarily to support messages sent during key transitions.

- **Security Improvements**
  - Sensitive request-body values, including device tokens, symmetric keys, and session identifiers, are now redacted from diagnostic logs.
  - Malformed payloads containing sensitive fields are handled conservatively to prevent accidental exposure.

### DIFF
--- a/apps/clientApp/src/androidMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/BisqFirebaseMessagingService.kt
+++ b/apps/clientApp/src/androidMain/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/BisqFirebaseMessagingService.kt
@@ -209,6 +209,11 @@ class BisqFirebaseMessagingService :
                     log.i { "Push decrypted with the previous key generation (sent before the last rotation)" }
                 }
                 return plaintext
+            } catch (e: kotlinx.coroutines.CancellationException) {
+                // Not a key mismatch: a cancelled caller must propagate, not fall through to
+                // the next candidate. Inert on the plain FCM callback thread, load-bearing the
+                // day this runs inside a coroutine.
+                throw e
             } catch (e: Exception) {
                 lastFailure = e
             }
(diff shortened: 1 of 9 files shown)


rodvar:
@coderabbitai full review

PR #1846, opened 2026-09-12
feat(kscan): replace ML Kit with zxing-cpp on Android

wodoro:
Progresses toward f-droid release preparations

Replaces mlkit with zxing-cpp on android. Frames are analyzed on a dedicated thread since zxing-cpp decodes synchronously; filter and result callbacks stay on the main thread.

ML Kit's zoom suggestions (auto-zoom) are removed. The manual inverted-frame pass is dropped in favour of tryInvert.

Final apk size reduction is expected. processes QRCodes up to 7 times faster.

## Summary by CodeRabbit

* **New Features**
  * Android barcode scanning now uses the open-source ZXing C++ engine instead of Google ML Kit, reducing reliance on Google Play Services.
  * Barcode detection supports ZXing symbologies and format variants.
  * Frame analysis runs on a dedicated background thread for improved responsiveness.
  * Android scanning now uses manual zoom controls; automatic zoom suggestions are no longer available.

* **Tests**
  * Added coverage for barcode format mapping and supported symbologies.

### DIFF
--- a/shared/kscan/README.md
+++ b/shared/kscan/README.md
@@ -8,3 +8,10 @@ Based on:
 
 - Commit: [754b88e8bf](https://github.com/ismai117/KScan/commit/754b88e8bf3841d608308d5bc3f97b7e208cea65)
 - Release tag: [0.9.2](https://github.com/ismai117/KScan/releases/tag/0.9.2)
+
+Differences from upstream:
+
+- Android decodes with [zxing-cpp](https://github.com/zxing-cpp/zxing-cpp) instead of ML Kit, so
+  every dependency is open source and no Google Play Services libraries are pulled in.
+- ML Kit's auto-zoom suggestions are removed. Zoom is manual only via `ScannerController` or the
+  built-in zoom controls.
--- a/shared/kscan/consumer-rules.pro
+++ b/shared/kscan/consumer-rules.pro
@@ -1,11 +1,8 @@
 # Consumer ProGuard rules for the KScan library
 # Packaged with the AAR and applied by consuming apps in release builds.
 
-# ML Kit vision barcode (keep internals referenced via reflection)
--keep class com.google.mlkit.** { *; }
--keep class com.google.android.gms.internal.mlkit_vision_barcode** { *; }
--dontwarn com.google.mlkit.**
--dontwarn com.google.android.gms.internal.mlkit_vision_barcode**
+# zxing-cpp ships its own keep rules for the classes its JNI layer reads, so
+# only CameraX needs naming here.
 
 # CameraX core and interop
 -keep class androidx.camera.** { *; }
@@ -13,4 +10,3 @@
 
 # Prevent obfuscation of KScan public API to keep binary compatibility
 -keep class org.ncgroup.kscan.** { *; }
-
--- a/shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/BarcodeAnalyzer.kt
+++ b/shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/BarcodeAnalyzer.kt
@@ -15,7 +15,7 @@ import java.util.concurrent.Executor
  * A barcode must be detected twice before it is reported, to filter out misreads.
  */
 class BarcodeAnalyzer(
-    codeTypes: List<BarcodeFormat>,
+    private val codeTypes: List<BarcodeFormat>,
     private val callbackExecutor: Executor,
     private val onSuccess: (List<Barcode>) -> Unit,
     private val onFailed: (Exception) -> Unit,
@@ -60,7 +60,8 @@ class BarcodeAnalyzer(
                 return
             }
 
-        val relevantResults = results.filter { BarcodeFormatMapper.isKnownFormat(it.format) }
+        val relevantResults =
+            results.filter { BarcodeFormatMapper.isRequested(BarcodeFormatMapper.toAppFormat(it.format), codeTypes) }
         if (relevantResults.isNotEmpty()) {
             callbackExecutor.execute { if (!closed) processFoundBarcodes(relevantResults) }
         }
@@ -82,7 +83,7 @@ class BarcodeAnalyzer(
                         rawBytes = rawBytes,
                     )
 
-                if (!filter(barcode)) return
+                if (!filter(barcode)) continue
 
                 hasSuccessfullyProcessedBarcode = true
                 barcodesDetected.clear()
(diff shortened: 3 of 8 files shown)


shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/BarcodeAnalyzer.kt
```diff
@@ -1,198 +1,102 @@
...
-    private fun scanInverted(imageProxy: ImageProxy) {
-        val invertedImage =
+        val results =
             try {
-                createInvertedInputImage(imageProxy)
+                imageProxy.use { reader.read(it) }
             } catch (e: Exception) {
-                // Conversion failed, clean up and exit
-                imageProxy.close()
+                callbackExecutor.execute { if (!closed) onFailed(e) }
```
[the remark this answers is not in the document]
wodoro: invalid since there's no coroutine or CancellationException involved?

shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/BarcodeFormatMapper.kt
```diff
@@ -1,60 +1,59 @@
...
-            return FORMAT_ALL_FORMATS
+            return allSupportedFormats
         }
-
-        return appFormats
-            .mapNotNull { APP_TO_MLKIT_FORMAT_MAP[it] }
-            .distinct()
-            .fold(0) { acc, formatInt -> acc or formatInt }
-            .let { if (it == 0) FORMAT_ALL_FORMATS else it }
+        return appFormats.mapNotNull { APP_TO_ZXING_FORMAT_MAP[it] }.toSet()
```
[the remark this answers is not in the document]
wodoro: addressed in  3b963a4b1

rodvar:
@coderabbitai full review

shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/ScannerView.android.kt
```diff
@@ -138,18 +142,18 @@ actual fun ScannerView(
                                 ).build(),
                         ).setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
                         .build()
+                        .also { imageAnalysis = it }
```
rodvar, line 145: nit - the local `val imageAnalysis` shadows the new state var, and this `.also` only assigns the outer one because it runs inside the initializer, before the local enters scope. Works, but easy to break in a refactor - maybe drop the local and use the state var directly, or rename one of them.

shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/BarcodeAnalyzer.kt
```diff
@@ -1,198 +1,103 @@
...
-            }
-    }
-
-    private fun scanInverted(imageProxy: ImageProxy) {
-        val invertedImage =
+        val results =
             try {
-                createInvertedInputImage(imageProxy)
+                imageProxy.use { reader.read(it) }
             } catch (e: Exception) {
```
rodvar, line 58: nit - a JNI-side `Error` would escape this catch and silently kill the executor task. Probably fine to leave as is, just flagging that the failure mode changed vs ML Kit's failure listener.

shared/kscan/src/androidMain/kotlin/org/ncgroup/kscan/BarcodeAnalyzer.kt
```diff
@@ -1,198 +1,103 @@
...
-            val rawBytes = mlKitBarcode.rawBytes ?: displayValue.encodeToByteArray()
+        for (result in results) {
+            val text = result.text ?: continue
+            val rawBytes = result.bytes ?: text.encodeToByteArray()
 
-            barcodesDetected[displayValue] = (barcodesDetected[displayValue] ?: 0) + 1
-            if ((barcodesDetected[displayValue] ?: 0) >= 2) {
-                val appSpecificFormat = BarcodeFormatMapper.toAppFormat(mlKitBarcode.format)
-                val detectedAppBarcode =
+            barcodesDetected[text] = (barcodesDetected[text] ?: 0) + 1
```
rodvar, line 77: nit - these counts persist across filter rejections and never expire. Same as before this PR and harmless for short scan sessions, just noting.

shared/kscan/README.md
```diff
@@ -8,3 +8,10 @@ Based on:
 
 - Commit: [754b88e8bf](https://github.com/ismai117/KScan/commit/754b88e8bf3841d608308d5bc3f97b7e208cea65)
 - Release tag: [0.9.2](https://github.com/ismai117/KScan/releases/tag/0.9.2)
+
+Differences from upstream:
+
+- Android decodes with [zxing-cpp](https://github.com/zxing-cpp/zxing-cpp) instead of ML Kit, so
+  every dependency is open source and no Google Play Services libraries are pulled in.
+- ML Kit's auto-zoom suggestions are removed. Zoom is manual only via `ScannerController` or the
```
rodvar, line 16: nit - with auto-zoom gone, worth a manual pass with a small/far QR to confirm manual zoom feels ok for the pairing flow. This is the only thing I haven't tested as its really minor.

shared/kscan/consumer-rules.pro
```diff
@@ -1,16 +1,12 @@
 # Consumer ProGuard rules for the KScan library
 # Packaged with the AAR and applied by consuming apps in release builds.
 
-# ML Kit vision barcode (keep internals referenced via reflection)
--keep class com.google.mlkit.** { *; }
--keep class com.google.android.gms.internal.mlkit_vision_barcode** { *; }
--dontwarn com.google.mlkit.**
--dontwarn com.google.android.gms.internal.mlkit_vision_barcode**
+# zxing-cpp ships its own keep rules for the classes its JNI layer reads, so
```
rodvar, line 4: checked the 3.1.1 aar and it does ship `-keep class zxingcpp.** { *; }`, so this holds. Will verify if tweaks are needed on the next release - this class of change only breaks under R8.

rodvar (approved):
ACK

Outstanding job, we don't get rid of kscan just adapt to remove the ML kit for QR code processing into https://github.com/zxing-cpp/zxing-cpp

Added some minor nits if agreed with any please send a follow up at your own time. 

Another nit that we are definitely accepting: on iOS the QR processing seems to be working slightly slower (is not as super instant as for Android - accepted)

Great job,
Thank you 🙏

shared/kscan/README.md
wodoro: in report for tests generated in https://github.com/ismai117/KScan/pull/61 , it seems it's very good at reading small QR codes:

> Small or distant symbols. With the symbol spanning 20% of the frame ML Kit
reads 61.5%, and at 12% it reads 23.1%; zxing-cpp reads 100% of both. This is
precisely the case autoZoom existed to paper over, so losing the parameter costs
less than it sounds.

Issue #135, opened 2024-12-20
Generic add to F-Droid request


ghost:

rodvar:
@endingisnight hey thanks for creating an issue proposing this! We are working hard on development of the MVP at the moment and adding to F-Droid was not part of the original effort estimation that the DAO approved, but obviously our apps are perfect candidates for it. We've added it to the next immediate milestone. Thanks 🙏

rodvar:
@endingisnight hope you are doing great. we might start working on this earlier than expected, could you recommend the best way to do it? 

I found an [old blogpost](https://kaeruct.github.io/posts/2021/08/03/publishing-an-app-on-f-droid/) of a successful experience, would that be a good tutorial to follow?

Thanks in advance! 🙏

rodvar:
So there are 2 ways of doing this, one is to host your own F-Droid server, with your apk and metadata 

## F-Droid self hosted

### Requirements

```
You have a good understanding of what [Free Software](https://www.gnu.org/philosophy/free-sw.html) means, and what F-Droid is for.
You already read and understand the [Inclusion Policy](https://f-droid.org/docs/Inclusion_Policy).
You already read and understand the [Repository Style Guide](https://f-droid.org/docs/Repository_Style_Guide).
You already read and understand [the relevant parts of the F-Droid documentation](https://f-droid.org/docs/Build_Metadata_Reference).
You know how to use [Git VCS](https://git-scm.com/), and know how a merge request (a.k.a. “pull request” in GitHub terminology) works in general.
You have an account on [GitLab](https://gitlab.com/) (NOTE: F-Droid CI runners are under Gitlab’s FOSS program, so there’s no need for you to pay for any CI time. If Gitlab starts asking for phone numbers or credit cards don’t submit anything, just leave a note in the MR so we know we need to trigger the CI.)
You have a local instance of the F-Droid server software, and you know what you are doing.
```

### How

 - Please refer to this [official guide](https://f-droid.org/docs/Setup_an_F-Droid_App_Repo/) , basically once you have the f-droid server on you need to lodge a [metadata merge request into their gitlab repo](https://gitlab.com/fdroid/fdroiddata/-/tree/master/metadata?ref_type=heads)

## F-Droid main public repo

### Requirements

 - apps need to be fully buildable from source (which we comply with already)
 - F-Droid team uses their own signature for the apk
 - they use apk not aab
 - a wait of ~2-3 weeks for review is anticipated.

### How

[...]

PR #1854, opened 2026-09-15
feat(clientApp): add fdroid flavor without FCM

wodoro:
- contribs to #135 

## Why

F-Droid's inclusion policy rejects proprietary dependencies outright rather than
flagging them as an anti-feature. `firebase-messaging`, plus the
`play-services-tasks` that `kotlinx-coroutines-play-services` pulls in, is one. So
Bisq Connect cannot ship on F-Droid as built today.

This adds an `fdroid` distribution flavor that carries no Google dependency at
all. `google` is unchanged and remains what ships to Play and as the sideloadable
GitHub APK.

### Flavor split

The two flavors differ in exactly one thing, the push transport:

|                     | `google`                              | `fdroid`                             |
| ------------------- | ------------------------------------- | ------------------------------------ |
| Relayed push        | FCM, opt-in, off by default           | none                                 |
| Background delivery | FCM plus the local foreground service | local foreground service only        |
| Settings opt-in     | shown                                 | shown, disabled, with an explanation |

`fdroid` binds an `UnsupportedPushNotificationTokenProvider`, whose
`isSupported = false` travels up through
`PushNotificationServiceFacade.isRelayedPushSupported` into
`SettingsUiState.isRelayedPushSupported`. The settings section still renders
there, with the switch disabled and an explanation replacing the usual tail text.
image
A setting that is simply absent reads as a bug to anyone who has seen it
documented. Behind it `ClientPushNotificationServiceFacade.activate()`
short-circuits, so nothing registers a device the trusted node could never reach.

The new string names no other build. F-Droid treats pointing users at a non-free
variant as promoting non-free software, so please keep that constraint in mind if
the wording is revised. It is noted in `docs/android.md` too.

### google-services scoping

Google's plugin creates a task per variant and resolves `google-services.json` for
all of them, with no per-variant switch, so it was generating `google_app_id` /
`google_api_key` / `gcm_defaultSenderId` string resources into the fdroid APK for
nothing to read. Non-Google variants now get an empty json list plus
`missingGoogleServicesStrategy = IGNORE`, so their task runs and produces nothing.

Disabling the task does not work: AGP merges its output folder whether or not it
ran, so a previous build's resources still reach the APK. Verified both ways with
a canary resource.

### Artifact naming

`AppArtifactsExtension` gains `primaryFlavor`. The named flavor keeps the
undecorated names so published release assets and the reproducible-build metadata
pinning them are unaffected; every other flavor gets its name appended last. Apps
without flavors take neither path.

```
google:  Bisq_Connect-<version>-<abi>-<versionCode>.apk
         BisqConnect-<version>_<versionCode>-<buildType>.aab
fdroid:  Bisq_Connect-<version>-<abi>-<versionCode>-fdroid.apk
         BisqConnect-<version>_<versionCode>-<buildType>-fdroid.aab
nodeApp: unchanged
```

APKs go through the existing rename task. AABs needed more: `archivesName` only
fixes the leading half of the name, and AGP appends `-<flavor>-<buildType>`
itself with no hook, so merely adding a second flavor would have renamed the Play
artifact. The plugin rewrites the BUNDLE artifact's output location to undo that.

### BUILD_TS replaced by BUILD_COMMIT (commit 2)

`BuildConfig.BUILD_TS` and `BuildNodeConfig.BUILD_TS` came from
`System.currentTimeMillis()`, so every build of either app was byte-different.
That was the one thing blocking bit-reproducible builds, which is what lets
F-Droid publish under the upstream signature instead of its own key.

`BUILD_COMMIT` serves the same identify-this-build purpose and is fixed for a
given source state. It is read via `providers.exec` so it stays a declared build
input: with the configuration cache on, reading git any other way bakes the hash
into the cache entry and later builds keep reporting a stale commit.
`MainPresenter` logs it beside the versions it already logs; `BUILD_TS` had no
readers at all.

### keystore-matrix script

`scripts/keystore-matrix/run-all.sh` hardcoded `apk/release` and `bundle/release`,
which no longer exist for a flavored app. Its path helpers now match at either
nesting depth, so clientApp's signing cases find their artifacts again, and they
now cover both flavors, which matches the script's own stated intent that every
shipped APK gets verified.

### CI workflow (commit 3)

`.github/workflows/fdroid.yml` builds what an F-Droid recipe produces, so the shape
is exercised here rather than discovered at submission. Two jobs, because their
costs differ by an order of magnitude:

- **client** builds the `fdroid` flavor directly.
- **node** builds bisq2 from source at the commit already pinned in
  `gradle/libs.versions.toml` and publishes the jars into the local Maven repo
  `settings.gradle.kts` already reads, then builds the node app. That is the
  expensive half, so it is cached on the pinned commit, with cache restore and save
  split and the save guarded on success: a run that dies midway through
  `publishAll` must not leave a half-populated `~/.m2` for later runs to trust.

Neither job needs signing secrets, so it runs unchanged on forks. It triggers on
`workflow_dispatch` (with an app selector and an ABI list) and on `connect_*` /
`anode_*` tags, each tag prefix waking only its own job.

Two steps go beyond plain packaging:

- the client job **fails if any fdroid APK contains** `com/google/firebase`,
  `com/google/android/gms` or the google-services string resources, so a dependency
  added later cannot quietly reintroduce Google code
- both jobs publish `sha256` sums, since the point of the F-Droid work is a
  reproducible build worth diffing against F-Droid's own output

## How to build the F-Droid artifacts

The intent is per-ABI APKs only for F-Droid: no universal APK, no AAB. That needs
no new build configuration, because `AppArtifactsPlugin` already supports it:

```bash
./gradlew :apps:clientApp:assembleFdroidRelease \
  -Pabi=armeabi-v7a,arm64-v8a,x86,x86_64
```

Naming an ABI set enables splits and drops the universal APK, and no bundle task
is invoked, so no AAB is produced. Verified on a debug build, which yields exactly
four APKs and nothing else:

```
Bisq_Connect-0.9.1-debug-armeabi-v7a-34001-fdroid.apk
Bisq_Connect-0.9.1-debug-arm64-v8a-34002-fdroid.apk
Bisq_Connect-0.9.1-debug-x86-34003-fdroid.apk
Bisq_Connect-0.9.1-debug-x86_64-34004-fdroid.apk
```

Version codes stay on the existing `base * 1000 + ABI ordinal` scheme, so they
interleave with the google flavor's rather than colliding with it.

## Release flow changes (clientApp only, nodeApp untouched)

clientApp now has flavors, so its task names and output paths are qualified. The
unqualified aggregates still exist but build _every_ flavor, which is not what a
Play release wants:

```
# before                             # now
:apps:clientApp:bundleRelease        :apps:clientApp:bundleGoogleRelease
:apps:clientApp:assembleRelease      :apps:clientApp:assembleGoogleRelease

apk/release/                         apk/google/release/
bundle/release/                      bundle/googleRelease/
```

nodeApp keeps `assembleRelease`, `bundleRelease`, `apk/release/` and
`bundle/release/` exactly as before. `docs/android.md` is updated with all of this.

## Verification

- fdroid APK dex: zero `com/google/firebase`, `com/google/android/gms` and FCM
  class references. google APK: 1110, unchanged.
- fdroid merged manifest: zero Firebase entries. google: all present.
- fdroid resources: none of the Google identifiers.
- Two clean builds of `:apps:clientApp:assembleFdroidDebug` produce byte-identical
  APKs.
- All six artifacts (APK and AAB, both flavors, plus nodeApp) built from a wiped
  outputs tree and checked against the names above.
- Both flavors' unit tests, `:shared:presentation` and `:shared:domain` tests, and
  `ktlintCheck` for both the main build and `build-logic` pass. nodeApp compiles.
- The workflow's client half was run locally end to end: `assembleFdroidRelease`
  with `-PallowUnsignedRelease=true -Pabi=...` yields exactly the per-ABI APKs and
  nothing else, they carry no APK signing block, and the Google-free guard passes
  against real R8-minified release APKs rather than only debug ones.

## Notes for reviewers

**Flavor sources use AGP's layout** (`src/google/`, `src/fdroid/`,
`src/testGoogle/`), not the `src/androidMain` spelling the rest of the module
uses. This is deliberate and worth knowing before moving anything: the KMP plugin
advertises an `src/androidGoogle` equivalent and does compile Kotlin from it, but
the variant manifest merger reads only AGP's path. A flavor manifest under the KMP
spelling is dropped with no error, which would ship a build that bundles FCM with
its service undeclared. The comment in `apps/clientApp/build.gradle.kts` records
it.

**Two ordering details are load-bearing** and commented in place, because both
fail silently if moved:

- the google-services task configuration sits in `afterEvaluate`, since the plugin
  wires the same properties from its own variant callback and would overwrite an
  earlier `configureEach`
- the AAB output location is set in a second `configure` block _after_
  `toTransform`, which appends its own action

**The workflow's bisq2 half has never been executed.** Its steps follow README.md
and the existing `release-umbrel-images.yml`, but nothing here could run
`publishAll` against a fresh clone, so treat the first run as the real test. Related:
bisq2 pins its own toolchain (`releaseBuild.javaVersion=21.0.11`, vendor Azul, with
`org.gradle.java.installations.auto-detect=false`). The workflow uses JDK 21 Zulu,
which matches that today, but the pin is not read from bisq2 and so is not enforced
if bisq2 later moves it. An F-Droid recipe has the same obligation, and in fdroiddata
that means selecting the JDK in the recipe's `sudo:` block.

**CI gets slower.** `:apps:clientApp:assembleDebug` is an aggregate that now builds
both flavors. I left it that way on purpose since it is the only thing giving the
fdroid variant compile coverage in CI. Narrowing it to `assembleGoogleDebug` is a
one-line change if the build time matters more than the drift protection.

## Not covered here

- **Analytics.** Sentry is declared in `shared/domain` commonMain, so it reaches
  both apps and iOS, and a clientApp product flavor cannot remove it. F-Droid would
  apply the Tracking anti-feature. The SDK is free software and the toggle is
  opt-in and default off, so accepting the tag is a legitimate option, but it needs
  a decision before submission. (Personally I think this is fine)
- **nodeApp needs no flavor.** Its only `com.google.*` dependencies are `protoc`
  and Guava, both free software. Shipping it on F-Droid is a packaging question
  (the bisq2 jars resolve from `mavenLocal` / a private Maven, so a recipe needs an
  F-Droid srclib at the pinned `bisq-core-commit`, which will be done by me at submission to f-droid), not a code change.
- **F-Droid metadata.** No `fdroiddata` recipe or fastlane metadata in this PR. (I will do a follow-up or I can leave this to you)
- kmp-tor and zxing-cpp ship prebuilt native libraries from Maven. Free software, both will be resolved at f-droid submission.

## New Features
* Added an F-Droid client distribution without Google services or Firebase.
* Added Google and F-Droid release variants with clearly labeled APK and bundle artifacts.
* Google builds continue to support optional relayed push notifications.
* Added automated F-Droid builds, APK verification, and checksum artifacts.

## Changes
* F-Droid builds identify relayed push notifications as unsupported.
* The push-notification setting remains visible but disabled with explanatory guidance.
* Background notifications remain available while the app is running.
* Build metadata now uses the source commit for reproducible releases.

## Documentation
* Updated Android build and release instructions for distribution variants.


apps/clientApp/src/fdroid/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/PushNotificationTokenProviderFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+ * dependencies outright. Background delivery here comes from the local foreground service holding
+ * the trusted-node WebSocket open instead.
+ *
+ * Reporting [isSupported] as false hides the settings opt-in (see
+ * `SettingsPresenter.shouldShowPushNotificationsToggle`), so these methods exist to keep the
+ * contract total rather than to be called. They still fail rather than fake success: a silent
+ * `Result.success` would register the device with the trusted node under a token nothing can
+ * deliver to, and the user would believe push is working.
+ */
+internal class UnsupportedPushNotificationTokenProvider : PushNotificationTokenProvider {
```
rodvar, line 23: please fix: this is where the PR Coverage gate fails (40% vs 80) - the whole flavor seam has no tests. A `src/testFdroid` unit test pinning this provider's contract (`isSupported == false`, `requestDeviceToken` fails rather than faking success) plus a facade-activate test with an unsupported provider (nothing registers, flows stay off) would cover it I think? 

And the never-fake-success property is load-bearing enough that it deserves to be pinned, not just documented. Might need kover variant wiring so the fdroid classes get instrumented instead of reported at 0%?

Also a nit while here --> the kdoc says the opt-in is "hidden" via `shouldShowPushNotificationsToggle`, but what ships is shown-disabled-with-explanation via `isRelayedPushSupported`. Looks like wording from an earlier iteration 🙏

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/push_notification/PushNotificationServiceFacade.kt
```diff
@@ -8,6 +8,15 @@ import network.bisq.mobile.data.service.LifeCycleAware
  * Handles device token registration with the trusted node.
  */
 interface PushNotificationServiceFacade : LifeCycleAware {
+    /**
+     * Whether this build can deliver relayed push notifications at all, independent of the
+     * user's opt-in. False where no transport is wired: the Node app, whose embedded Bisq2
+     * process notifies locally, and Connect's fdroid flavor, which ships without FCM because
+     * F-Droid rejects proprietary dependencies. The settings opt-in is hidden when this is
```
rodvar: nit - "the settings opt-in is hidden when this is false" is stale: on the fdroid flavor the section renders with the switch disabled and an explanation (which the SettingsPresenter kdoc describes correctly); only the Node app hides it, via shouldShowPushNotificationsToggle. Same wording survives in `PushNotificationTokenProvider.isSupported`'s kdoc.

docs/android.md
```diff
@@ -7,25 +7,92 @@ a dependency or an API call that assumes a modern JDK.
...
+`missingGoogleServicesStrategy = IGNORE`, so their task runs and produces nothing. Without it an
+fdroid APK built on a machine that has the json picks up `google_app_id` / `google_api_key` /
+`gcm_defaultSenderId` string resources that nothing reads. Disabling the task instead does not
+work: AGP merges its output folder whether or not it ran, so the previous build's resources would
+still reach the APK.
+
+Artifact names: `google` APKs keep the undecorated `Bisq_Connect-<version>-<abi>-<versionCode>.apk`
+so published assets and the reproducible-build metadata pinning them are unaffected, and fdroid
+APKs carry the flavor as a trailing segment:
+`Bisq_Connect-<version>-<abi>-<versionCode>-fdroid.apk`. That split is
```
rodvar, line 96: question - two things worth a sentence here: (1) users can never cross-grade a google install to an fdroid one or back (different signing certs), and (2) the two flavors deliberately share versionCodes, which is fine because F-Droid signs and distributes independently - stating it stops someone from "fixing" it later.

rodvar:
@coderabbitai full review

rodvar (changes requested):
Thanks for this PR with the extensive documentation, it was very helpful to digest how this was done and the approach and decisions made are exactly what I was expecting - well done wod.

I did a first code review pass on this one before fully testing, one major issue (coverage) and all the rest are nits. 

Please let's agree and address this first before I get hands on fully testing it. Thank you 💪

rodvar:
@wodoro let's please also review the CRAI comments above 🙏

shared/domain/src/commonMain/kotlin/network/bisq/mobile/data/service/push_notification/PushNotificationServiceFacade.kt
wodoro: correct, at first it was hidden, but I decided to show and disable it and had iterated, but forgot to check for stale comments

wodoro:
@coderabbitai continue

apps/clientApp/src/google/kotlin/network/bisq/mobile/client/common/domain/service/push_notification/PushNotificationTokenProviderFactory.kt
```diff
@@ -0,0 +1,8 @@
+package network.bisq.mobile.client.common.domain.service.push_notification
+
+/**
+ * Flavor seam for the relayed-push transport, declared once per distribution flavor.
+ * Duplicated rather than expressed as expect/actual because those span Kotlin targets,
+ * not Android product flavors.
+ */
+internal fun createPushNotificationTokenProvider(): PushNotificationTokenProvider = AndroidPushNotificationTokenProvider()
```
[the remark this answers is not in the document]
wodoro: validate again

scripts/verify-fdroid-apk.py
```diff
@@ -0,0 +1,110 @@
...
+
+import collections
+import re
+import sys
+import zipfile
+
+# Class references appear in dex as type descriptors, `Lcom/google/common/base/Strings;`.
+# Anchoring on the leading `L` keeps incidental strings (protobuf descriptor data names a
+# few packages it does not ship) from being read as shipped classes.
+CLASS_REF = re.compile(rb"Lcom/google/([a-zA-Z0-9_]+)")
```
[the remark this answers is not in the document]
wodoro: validate again

rodvar:
@coderabbitai continue

docs/android.md
```diff
@@ -7,25 +7,126 @@ a dependency or an API call that assumes a modern JDK.
...
 
 ```bash
-./gradlew apps:[type]:clean apps:[type]:bundleRelease --info && ./gradlew apps:[type]:assembleRelease --info
+# Bisq Connect (clientApp). `google` is the flavor that ships to Play and GitHub.
+./gradlew apps:clientApp:clean apps:clientApp:bundleGoogleRelease --info && ./gradlew apps:clientApp:assembleGoogleRelease --info
+
+# Bisq Connect, fdroid flavor. Per-ABI only: F-Droid takes no AAB, and naming the ABIs also
+# drops the universal APK. Signed like everything else, because F-Droid's reproducible-builds
+# path verifies its own rebuild against these and then redistributes them.
+./gradlew apps:clientApp:assembleFdroidRelease -Pabi=armeabi-v7a,arm64-v8a,x86,x86_64 --info
```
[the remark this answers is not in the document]
wodoro: verify again

scripts/verify-fdroid-apk.py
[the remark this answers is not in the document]
wodoro: verify again

rodvar (approved):
~~ACK~~

~~please review my pre-merge commit. We need to improve a bit the vertical positioning / padding further, follow up in a different PR.~~

~~Thank you buddha~~ 🙏

rodvar (changes requested):
sorry last comment was for a different review, haven't checked this one yet

apps/clientApp/src/google/AndroidManifest.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
```
rodvar, line 1: 👏

shared/domain/src/commonMain/resources/mobile/mobile.properties
```diff
@@ -173,6 +173,7 @@ mobile.pushNotifications.settings.subtitle=Receive trade alerts when the app is
 mobile.pushNotifications.settings.iosOnlyOption=On iOS, this is the only way to receive notifications when the app is not running.
 mobile.pushNotifications.settings.learnMoreLink=How this works and what is shared →
 mobile.pushNotifications.settings.disabledWarning=You may miss trade messages while the app is closed.
+mobile.pushNotifications.settings.unsupportedOnThisBuild=Not available in this build, which ships without Google services. Notifications still arrive while Bisq keeps running in the background.
```
rodvar, line 176: 👏

shared/presentation/src/commonMain/kotlin/network/bisq/mobile/presentation/main/MainPresenter.kt
```diff
@@ -131,6 +131,7 @@ open class MainPresenter(
         val screenWidth = getScreenWidthDp()
         _isSmallScreen.value = screenWidth < 480
         log.i { "Shared Version: ${BuildConfig.SHARED_LIBS_VERSION}" }
+        log.i { "Build Commit: ${BuildConfig.BUILD_COMMIT}" }
```
rodvar, line 134: 👏

.github/workflows/fdroid.yml
```diff
@@ -0,0 +1,341 @@
+name: F-Droid artifacts
```
rodvar, line 1: 👍

rodvar (approved):
ACK

Outstanding work @wodoro 

This PR was a pleasure to read and review. All the concerns above have been addressed.

The implementation for this is clear, well documented and tested, and the UX decisions are excellent IMHO. 

Brilliant work.

Some notes for this and future work:

 1. nit: as discussed I think its better to not marry the de-googled builds to `fdroid` in the build's naming conventions. But we agreed its ok to keep for not and generalise later if we have more sideloding appstores
 2. I see the fdroid new CI workflow generates builds. Bear in mind those will be kept for ~14 days I think in GH, so I understand we do it for verification purposes only.
 3. Although I'm merging this now (before the release of nodeApp today), this week release won't include the f-droid apks 
 4. I believe, to help me with Release Management, some follow ups are needed (see below) - please let me know to what extend would you like to take them or if you prefer me to do it

```

  - Make a release script that owns F-Droid handling, not CI (verification). The workflow stays a canary: unsigned builds, digests, no release upload.
  - F-Droid publishing needs an fdroiddata recipe outside this repo, and F-Droid's builder pulls tags itself. (prob wod already doing this?)
  - The signing choice is still open and sticky. F-Droid's own key means no extra release steps. The reproducible path means the script stages the signed fdroid APKs into the GitHub release asset set, and the docs paragraph in android.md must match whichever you pick. (please let's chat about this on matrix wod?)

```

Thank you!!! 💪 💪

wodoro:
Regarding f-droid publishing: yes, I will submit a MR to the f-droid data repo with the correct build metadata file when everything is in place

PR #1865, opened 2026-09-18
fix(build): fail closed on unsigned flavored releases

wodoro:
## Why

Two problems in `gradle/releaseSigning.gradle.kts`, both surfaced while reviewing what an
F-Droid recipe would need.

**1. The fail-closed keystore check stopped running for Bisq Connect.**

The check matched four literal task names: `assembleRelease`, `bundleRelease`,
`packageRelease`, `packageReleaseBundle`. As soon as clientApp gained the
`google` / `fdroid` flavor split, AGP started folding the flavor into those names, so the
app now packages through `packageFdroidRelease` and `bundleGoogleRelease` and the graph
never contains any of the four. A `--dry-run` of `:apps:clientApp:assembleFdroidRelease`
confirms it: the graph holds `packageFdroidRelease` and `assembleFdroidRelease`, nothing
else.

Result: a missing or mistyped `KEYSTORE_PATH` silently produced an **unsigned** release
APK instead of failing. That applies to the documented Play command
(`bundleGoogleRelease`) as much as to the F-Droid one. It matters more than usual here
because the F-Droid plan is the reproducible-builds path, where F-Droid pins our signing
key via `AllowedAPKSigningKeys` and verifies its rebuild against the APK we publish. An
unsigned reference binary breaks that quietly.

`nodeApp` was never affected, having no flavors.

**2. `local.properties` was mandatory to configure the build.**

`load(file("${rootDir}/local.properties").inputStream())` ran unconditionally, so a clean
checkout failed at configuration time. Both workflows worked around it by writing a dummy
file with `KEYSTORE_PASSWORD` and the key aliases but no `KEYSTORE_PATH`.

None of those values were ever read. The companion properties are required only inside
`if (releaseKeystoreFile != null)`, and with no `KEYSTORE_PATH` there is no keystore, no
`signingConfigs.release`, and nothing to validate. The file existed purely to stop the
`load()` call from throwing.

The actual opt-in for a keystore-less release already exists and is already used by these
workflows: `-PallowUnsignedRelease=true`, checked by
`check(releaseKeystoreFile != null || allowUnsignedRelease)`. The dummy file was a
workaround for an I/O bug dressed up as signing configuration, which is why it looked
load-bearing.

## What changed

- `gradle/releaseSigning.gradle.kts`
  - `local.properties` is loaded only when present, through `use {}` so the stream closes.
  - Release and profile packaging tasks are matched by pattern rather than by a fixed list:
    `^(assemble|bundle|package)([A-Z][A-Za-z0-9]*)?Release(Bundle)?$` and the `Profile`
    equivalent. Anchoring on the build type keeps `packageReleaseResources` and the
    `*ReleaseUnitTest` tasks out. The two `allTasks.any` blocks collapse into one helper.
- `.github/workflows/fdroid.yml` — dummy `local.properties` steps removed from both jobs, so
  the client job now exercises the same configuration an F-Droid recipe would, from a bare
  checkout.
- `.github/workflows/ci.yml` — same three steps removed. No job here packages a release, and
  the SDK location comes from `setup-android`, not `sdk.dir`.

## Verification

| Command | Result |
|---|---|
| `:apps:clientApp:assembleFdroidRelease --dry-run` | fails with `Release packaging needs a readable KEYSTORE_PATH, or pass -PallowUnsignedRelease=true for an unsigned APK` |
| same, `-PallowUnsignedRelease=true` | `BUILD SUCCESSFUL` |
| `:apps:clientApp:bundleGoogleRelease --dry-run` | now guarded (was not before) |
| `:apps:nodeApp:assembleRelease --dry-run` | still guarded, no regression |
| `:apps:nodeApp:testDebugUnitTest :apps:nodeApp:testReleaseUnitTest --dry-run` | `BUILD SUCCESSFUL` |
| `:apps:clientApp:assembleDebug --dry-run` | `BUILD SUCCESSFUL` |
| `ktlintCheck -Pkotlin.native.ignoreDisabledTargets=true` (real run) | `BUILD SUCCESSFUL`, 280 tasks |

The first row proves both fixes at once: configuration now succeeds without
`local.properties`, and the guard fires on a flavored task.

## Behaviour change

A developer with no `KEYSTORE_PATH` who runs `assembleFdroidRelease` or
`assembleGoogleRelease` now gets an error instead of an unsigned APK. That is the intent;
`-PallowUnsignedRelease=true` is the way through, and `docs/android.md` already assumes a
configured keystore for real releases.

## Summary by CodeRabbit

* **Chores**
  * Streamlined automated build and verification workflows by removing unnecessary placeholder configuration.
  * Improved build compatibility for different release variants, including F-Droid and Google release packages.
  * Release configuration now handles clean environments without requiring local signing properties.


rodvar (approved):
ACK

nice catch, verified the old list matched zero clientApp tasks on the documented commands.

Some follow ups suggested:

 - **please fix:** keystore-matrix only runs the unflavored aggregates so it would never have caught this, add bundleGoogleRelease / assembleFdroidRelease -Pabi cases (follow-up is accepted)
 - **nit** - the reproducible-builds path isn't decided yet, code is neutral so ok, just don't take the description as the decision

